PR #17, opened 2020-11-26
Goal: full compatibility with gym environments through the rlberry.wrappers.Wrapper class.

omardrwch:
Wrapper can now handle any kind of `gym` space. `rlberry` spaces implement custom reseed() and sample() methods, when compared to `gym`.

This is to ensure that `rlberry` can control the random number generators of all objects that it handles through its unified seeding mechanism. 

Included `gym` as requirement in `setup.py`.


eleurent (approved):
One comment related to a sentence in the commit message "This is to ensure that rlberry can control the random number generators of all objects that it handles through its unified seeding mechanism." -> rlberry still won't handle the external envs' internal rngs (typically used in `env.step(action)`), right?

omardrwch:
> One comment related to a sentence in the commit message "This is to ensure that rlberry can control the random number generators of all objects that it handles through its unified seeding mechanism." -> rlberry still won't handle the external envs' internal rngs (typically used in `env.step(action)`), right?

Right, but calling `wrapped_env.reseed()` will do the following:
* Generate a random integer N using a `rlberry` random number generator
* Call  `wrapped_env.env.seed(N)` in the wrapped `gym` environment.

So I believe this should ensure reproducibility when using gym environments.

yfletberliac (approved):

eleurent:
@omardrwch But then, couldn't the same idea be used to reseed the env's gym spaces through their [seed](https://github.com/openai/gym/blob/master/gym/spaces/space.py#L39) methods? (this could be done in `Wrapper.reseed()`)

In which case, there is no need to wrap them, and redefine their sample() method etc.

eleurent (changes requested):
@omardrwch But then, couldn't the same idea be used to reseed the env's gym spaces through their [seed](https://github.com/openai/gym/blob/master/gym/spaces/space.py#L39) methods? (this could be done in `Wrapper.reseed()`)

In which case, there is no need to wrap them, and redefine their sample() method etc.

omardrwch:
@eleurent I don't think reseeding works very well in gym spaces, e.g., we don't seem to be able to reseed them twice:

```
    def np_random(self):
        """Lazily seed the rng since this is expensive and only needed if
        sampling from this space.
        """
        if self._np_random is None:
            self.seed()

        return self._np_random

    def seed(self, seed=None):
        """Seed the PRNG of this space. """
        self._np_random, seed = seeding.np_random(seed)
        return [seed]
```

Source: https://github.com/openai/gym/blob/master/gym/spaces/space.py

eleurent:
> we don't seem to be able to reseed them twice

Why, yes of course we can reseed them twice. I don't see what you mean.
The `seed()` method can be called multiple time, and updates the internal rng with the proper seed.
the `np_random()` getter is only there to provide an rng (with a random seed) when the space has not been seeded yet, but it gets overriden at further calls of `seed()`.

omardrwch:
@eleurent Ah yes, indeed! I got confused with these two `np_random()`. I'll implement what you suggested.

eleurent (approved):

PR #25, opened 2020-11-27
Add the DQN algorithm

### LINKED ISSUE #19: Add DQN

eleurent:
This PR adds the implementation of the DQN algorithm, with the following features:

- Abstract algorithm description + pytorch implementation
- Automatic selection of the best GPU available (with most free memory), or switch to CPU when GPUs are unavailable
- Use Double DQN trick
- Support for various configurable NN models (MLP, CNN, Self-Attention) + Dueling architecture.
- Visualisation of training progress with tensorboard
- Working example script on CartPole
- Unit test

Fixes #19

### DIFF
--- a/rlberry/agents/common/models.py
+++ b/rlberry/agents/common/models.py
@@ -140,7 +140,8 @@ class MultiLayerPerceptron(BaseModule):
         self.layer_sizes = layer_sizes or [64, 64]
         self.out_size = out_size
         sizes = [in_size] + self.layer_sizes
-        layers_list = [nn.Linear(sizes[i], sizes[i + 1]) for i in range(len(sizes) - 1)]
+        layers_list = [nn.Linear(sizes[i], sizes[i + 1])
+                       for i in range(len(sizes) - 1)]
         self.layers = nn.ModuleList(layers_list)
         if out_size:
             self.predict = nn.Linear(sizes[-1], out_size)
@@ -180,7 +181,8 @@ class DuelingNetwork(BaseModule):
         x = self.base_module(x)
         value = self.value(x).expand(-1, self.out_size)
         advantage = self.advantage(x)
-        return value + advantage - advantage.mean(1).unsqueeze(1).expand(-1, self.out_size)
+        return value + advantage \
+            - advantage.mean(1).unsqueeze(1).expand(-1, self.out_size)
 
 
 class ConvolutionalNetwork(nn.Module):
@@ -202,7 +204,7 @@ class ConvolutionalNetwork(nn.Module):
         # Number of Linear input connections depends on output of conv2d layers
         # and therefore the input image size, so compute it.
         def conv2d_size_out(size, kernel_size=2, stride=2):
-            return (size - (kernel_size - 1) - 1) // stride  + 1
+            return (size - (kernel_size - 1) - 1) // stride + 1
         convw = conv2d_size_out(conv2d_size_out(conv2d_size_out(in_width)))
         convh = conv2d_size_out(conv2d_size_out(conv2d_size_out(in_height)))
         assert convh > 0 and convw > 0
@@ -233,10 +235,18 @@ class EgoAttention(BaseModule):
         self.dropout_factor = dropout_factor
         self.features_per_head = int(self.feature_size / self.heads)
 
-        self.value_all = nn.Linear(self.feature_size, self.feature_size, bias=False)
-        self.key_all = nn.Linear(self.feature_size, self.feature_size, bias=False)
-        self.query_ego = nn.Linear(self.feature_size, self.feature_size, bias=False)
-        self.attention_combine = nn.Linear(self.feature_size, self.feature_size, bias=False)
+        self.value_all = nn.Linear(self.feature_size,
+                                   self.feature_size,
+                                   bias=False)
+        self.key_all = nn.Linear(self.feature_size,
+                                 self.feature_size,
+                                 bias=False)
+        self.query_ego = nn.Linear(self.feature_size,
+                                   self.feature_size,
+                                   bias=False)
+        self.attention_combine = nn.Linear(self.feature_size,
+                                           self.feature_size,
+                                           bias=False)
 
     @classmethod
     def default_config(cls):
@@ -246,21 +256,36 @@ class EgoAttention(BaseModule):
     def forward(self, ego, others, mask=None):
         batch_size = others.shape[0]
         n_entities = others.shape[1] + 1
-        input_all = torch.cat((ego.view(batch_size, 1, self.feature_size), others), dim=1)
+        input_all = torch.cat((ego.view(batch_size, 1,
+                               self.feature_size), others), dim=1)
         # Dimensions: Batch, entity, head, feature_per_head
-        key_all = self.key_all(input_all).view(batch_size, n_entities, self.heads, self.features_per_head)
-        value_all = self.value_all(input_all).view(batch_size, n_entities, self.heads, self.features_per_head)
-        query_ego = self.query_ego(ego).view(batch_size, 1, self.heads, self.features_per_head)
+        key_all = self.key_all(input_all).view(batch_size,
+                                               n_entities,
+                                               self.heads,
+                                               self.features_per_head)
+        value_all = self.value_all(input_all).view(batch_size,
+                                                   n_entities,
+                                                   self.heads,
+                                                   self.features_per_head)
+        query_ego = self.query_ego(ego).view(batch_size, 1,
+                                             self.heads,
+                                             self.features_per_head)
 
         # Dimensions: Batch, head, entity, feature_per_head
         key_all = key_all.permute(0, 2, 1, 3)
         value_all = value_all.permute(0, 2, 1, 3)
         query_ego = query_ego.permute(0, 2, 1, 3)
         if mask is not None:
-            mask = mask.view((batch_size, 1, 1, n_entities)).repeat((1, self.heads, 1, 1))
-        value, attention_matrix = attention(query_ego, key_all, value_all, mask,
+            mask = mask.view((batch_size, 1, 1,
+                              n_entities)).repeat((1, self.heads, 1, 1))
+        value, attention_matrix = attention(query_ego,
+                                            key_all,
+                                            value_all,
+                                            mask,
                                             nn.Dropout(self.dropout_factor))
-        result = (self.attention_combine(value.reshape((batch_size, self.feature_size))) + ego.squeeze(1))/2
+        result = (self.attention_combine(
+                    value.reshape((batch_size,
+                                   self.feature_size))) + ego.squeeze(1))/2
         return result, attention_matrix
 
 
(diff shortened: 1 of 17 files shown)


rlberry/agents/common/models.py
```diff
@@ -0,0 +1,524 @@
...
+
+        action_logprobs = dist.log_prob(action)
+        dist_entropy = dist.entropy()
+
+        state_value = self.critic(state)
+
+        return action_logprobs, torch.squeeze(state_value), dist_entropy
+
+
+class Net(nn.Module):
```
yfletberliac, line 94: Isn't this a duplicate of the class `MultiLayerPerceptron` ?

eleurent: Not exactly: MultiLayerPerceptron is more configurable (number of layers, type of activation, etc.) while Net is forced to have 2-layers Relu.

However, Net is used by your policy gradient algorithms, and I prefer keeping it as it is for this PR, so that the changes are only related to DQN with minimal modifications to other parts of the library (i.e. PG algorithms).
I suggest that in a future commit, we improve code sharing by removing Net and replace it with MultiLayerPerceptron everywhere it is used. 

The same must be done for the `Memory` classes, which is reimplemented by each algorithm (A2C, PPO, AVEC PPO, CEM, and DQN) with very similar code, while it should be common code instead. But let's merge them independently from this DQN feature.

rlberry/agents/dqn/pytorch.py
```diff
@@ -0,0 +1,104 @@
...
+from rlberry.agents.common.training import loss_function_factory, optimizer_factory
+from rlberry.agents.dqn.abstract import AbstractDQNAgent
+from rlberry.agents.dqn.memory import Transition
+from rlberry.utils.torch import choose_device
+
+logger = logging.getLogger(__name__)
+
+
+class DQNAgent(AbstractDQNAgent):
+    def __init__(self, env, config=None):
```
omardrwch, line 15: For compatibility with the `rlberry.agents.Agent` class, I think this line should be:

```
def __init__(self, env, **config):
```

and I believe the rest of the code can remain unchanged, right?

rlberry/agents/dqn/abstract.py
```diff
@@ -0,0 +1,179 @@
...
+
+from rlberry.agents import Agent
+from rlberry.agents.dqn.exploration import exploration_factory
+from rlberry.agents.dqn.memory import ReplayMemory, Transition
+from rlberry.utils.configuration import Configurable
+
+
+class AbstractDQNAgent(Configurable, Agent, ABC):
+    def __init__(self, env, config=None):
+        Agent.__init__(self, env, copy_env=False, reseed_env=False)
```
omardrwch, line 12:
```suggestion
class AbstractDQNAgent(Configurable, Agent, ABC):
    def __init__(self, env, **config):
        Agent.__init__(self, env, **config)
```

rlberry/agents/tests/test_dqn.py
```diff
@@ -0,0 +1,10 @@
+import gym
+
+from rlberry.agents.dqn.pytorch import DQNAgent
+
+
+def test_dqn_agent():
+    env = gym.make("CartPole-v0")
```
omardrwch, line 7:
```suggestion
import gym
from rlberry.wrappers import Wrapper
from rlberry.agents.dqn.pytorch import DQNAgent


def test_dqn_agent():
    env = Wrapper(gym.make("CartPole-v0"))
```

rlberry/agents/dqn/pytorch.py
eleurent: Yes, that is inherited from my personal way of handling configurations: a Configurable object should have all its config in a `config` dict, so that it can be saved to and loaded from a text file, in a generic way. See an example configuration here: https://github.com/eleurent/rl-agents/blob/master/scripts/configs/HighwayEnv/agents/DQNAgent/dqn.json

I'll see if I can adapt it to the current convention. But I don't really see how we can distinguish configurations (which are relevant to save/load and edit in a text format) from other run-specific variables (current episode, rewards, etc.). Using pickle does not seem very advisable here.

omardrwch: What about doing the following:

```python
class SomeAgent(Agent, Configurable):
     def __init__(self, env, **config):
            Agent.__init__(self, env, **config)
            Configurable.__init__(self, config)
```

eleurent: Interesting, that might work!
But I don't think it's a good idea to keep both conventions, I'll try to use yours.

Pros:
+ I recognize that it has the added benefit that the agent configuration are more visible in the constructor interface, where they probably should be.
+ type hinting can be used

Cons:
- mutables cannot be used as default values for configurations, as they could when defining a config as a dict.
e.g. `class Network(layers=[10, 10, 10])` is forbidden, while default_config = {"layers": [10, 10, 10]} is allowed

rlberry/agents/dqn/pytorch.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def initialize_model(self):
+        self.value_net.reset()
+
+    def set_writer(self, writer):
+        super().set_writer(writer)
+        obs_shape = self.env.observation_space.shape if isinstance(self.env.observation_space, spaces.Box) else \
+            self.env.observation_space.spaces[0].shape
+        model_input = torch.zeros((1, *obs_shape), dtype=torch.float, device=self.device)
+        self.writer.add_graph(self.value_net, input_to_model=(model_input,)),
```
xuedong, line 103: Why there is a , at the end of this line?

rlberry/agents/dqn/pytorch.py
omardrwch: I see, I think we can do the following for the default mutables:

```python
class Network:
    def __init__(self, layers=None):
           if layers is None:
              layers = [10, 10, 10]
```
or something like this. Then, we can put the default values in the docstring. Some editors, as vscode, show the dosctring and the required parameters by the class as we write the code. So this approach should overall be more user-friendly: 

image

eleurent: Yes, but then the default configuration is more buried and part of the code, rather than at the interface or in a separate dict. Using the doc is perhaps a good idea, but harder to maintain.

But I agree that it is the intended python syntax, so that's a very compelling argument.

My main concern is the ability to save/load agent configurations, with their hierarchical structure, in a text format that can be easily edited.

omardrwch: Do we really need this standardised way to save/load agent configurations?

Why can't we just write a config.py file, defining all the parameters/dictionaries required by the agent? 
Then, if we want to keep a backup of the parameters of an experiment, the agent can just copy this file to some output folder.

I like this approach because it makes the code much simpler for the user, and:

* It seems to avoid the issues in the discussion, and makes it easy to have a "transparent" constructor, where we can see all the parameters, instead of passing dictionaries everywhere.

* It is much more flexible and avoids complicated dictionaries as input. For example, I'm writing a code for an agent that takes an UncertaintyEstimator (see rlberry/exploration_tools) as input, and I'm doing the following:
```python
class AgentWithBonus:
    def __init__(self, env, n_episodes, gamma, uncertainty_estimator_fn):
          self.uncertainty_estimator = uncertainty_estimator_fn()
         ...
```
where `uncertainty_estimator_fn` is a function returning one instance of an UncertaintyEstimator.  To create such instance, we need more parameters. 
Writing the constructor as I did above, simplifies a lot the Agent code: we don't need to send an extra dictionary for the agent, so that it creates an instance UncertaintyEstimator using this dictionary.

Then, we can define `uncertainty_estimator_fn` in a `config.py` file easily. 

I'm using the same strategy for algorithms using feature maps, where I send a function `feature_map_fn` to the agent constructor, that can create an instance of a `FeatureMap`.

We could do the same for creating instances of neural nets too.

Notice that this is not incompatible with using `.json` config files, the user can always make a `config.py` that reads a `config.json`, and, after the experiments, save this `config.json` elsewhere as backup.

omardrwch: To summarize the above:

I don't think the agent should handle all the parameters of everything it creates. In the example I used, `UncertaintyEstimator` can be a lot of things that implement the same interface: a class that counts the number of visits to each (s, a);  a random network distillation, a density estimator etc.  By doing what I said, the agent doesn't need to worry about this.

eleurent: >Why can't we just write a config.py file, defining all the parameters/dictionaries required by the agent?
> Then, if we want to keep a backup of the parameters of an experiment, the agent can just copy this file to some output folder.

It just seems like a very bad practice to me, I think there's a reason why yml files are used everywhere for configurations. The python code used to store config values in your proposition makes it prone to side effects (no boundary over what this code does), cannot be parsed by another module, compared to another configuration, etc.

> It seems to avoid the issues in the discussion, and makes it easy to have a "transparent" constructor, where we can see all the parameters, instead of passing dictionaries everywhere.

Yes, I agree with this.

> Writing the constructor as I did above, simplifies a lot the Agent code: we don't need to send an extra dictionary to the agent, so that it creates an instance UncertaintyEstimator using this dictionary. Then, we can define uncertainty_estimator_fn in a config.py file easily.

I don't see what using a function brings, you could've sent the constructed UncertaintyEstimator as a parameter directly. But then, you are only breaking the construction of the agent into many separate parts, which seems to make it more complicated for the user, compared to just one call to the Agent contructor.

> Regarding .json config files, the user can always make a config.py that reads a config.json, and, after the experiments, save this config.json elsewhere as backup

Having a reliable and flexible configuration system seems an important feature to me, especially given how hard it is to debug/reproduce RL results. It should not be the responsibility of the user to implement it.

> I don't think the agent should handle all the parameters of everything it creates. In the example I used, UncertaintyEstimator can be a lot of things that implement the same interface: a class that counts the number of visits to each (s, a); a random network distillation, a density estimator etc. By doing what I said, the agent doesn't need to worry about this.

I absolutely agree, and it is also the case if the agent merely forwards to the UncertaintyEstimator (or any subpart) its configuration (e.g.), and it is the UncertaintyEstimator's responsibility to decode it (and fill in missing values using defaults).

eleurent:
Thanks for fixing the coding style!

rlberry/agents/dqn/pytorch.py
omardrwch: > It just seems like a very bad practice to me, I think there's a reason why yml files are used everywhere for configurations. The python code used to store config values in your proposition makes it prone to side effects (no boundary over what this code does), cannot be parsed by another module, compared to another configuration, etc.
[...]
> Having a reliable and flexible configuration system seems an important feature to me, especially given how hard it is to debug/reproduce RL results. It should not be the responsibility of the user to implement it.

I agree! A friend recommended the [Sacred](https://github.com/IDSIA/sacred) library, and it can be easily be included in the pipeline as in [this notebook that I added to the examples](https://colab.research.google.com/github/rlberry-py/rlberry/blob/main/notebooks/experimental_pipeline_with_rlberry.ipynb).
`Sacred` can read/write  .json or .yml files to reproduce the experiments and it's very easy to use with `rlberry`.
What to you think ?

> I don't see what using a function brings, you could've sent the constructed UncertaintyEstimator as a parameter directly. But then, you are only breaking the construction of the agent into many separate parts, which seems to make it more complicated for the user, compared to just one call to the Agent contructor.

If we are creating several agents, passing the a constructed UncertaintyEstimator directly would require us to construct several instances (by deepcopy, for example) to send to each agent. So passing a function solves this issue. 

I think it's simpler to pass a function to the constructor than passing nested dictionaries. Also, by doing that, the agent does not need to know which subclass of UncertaintyEstimator it has to instantiate.

PR #43, opened 2021-02-16
Adding seeding in ppo.py to enforce seeding (seeding.set_global_seed not enough)

yfletberliac:
Hello,

Just a very short PR because I would like to see with you what you think is the best way to seed a call to `np.random.choice` inside PPO.

I'm using `from rlberry.seeding import seeding` in `agents/ppo/ppos.py`.

In fact, I wonder why this is not already taken care of by `seeding.set_global_seed(1223)` in `examples/demo_ppo.py`. If you try running `examples/demo_ppo.py` without the change of this PR, `rd_indices` L304 will produce different integers each time (hence I label the PR as Bug, but I may be wrong here).

What do you think?

### DIFF
--- a/rlberry/agents/ppo/ppo.py
+++ b/rlberry/agents/ppo/ppo.py
@@ -149,6 +149,7 @@ class PPOAgent(IncrementalAgent):
         assert isinstance(self.env.action_space, spaces.Discrete)
 
         self.cat_policy = None  # categorical policy function
+        self.rng = seeding.get_rng()
 
         # initialize
         self.reset()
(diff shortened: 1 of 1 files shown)


omardrwch (approved):

rlberry/agents/ppo/ppo.py
```diff
@@ -299,7 +300,8 @@ def _update(self):
         for _ in range(self.k_epochs):
 
             # shuffle samples
-            rd_indices = np.random.choice(n_samples, size=n_samples, replace=False)
+            np_random = seeding.get_rng()
```
omardrwch, line 303: `get_rng()` should be in `__init__()`, the agent should have a single `rng` object:

```python
self.rng = seeding.get_rng()   # in __init__()
# ...
self.rng.choice(...)   # where you need it
```

PR #45, opened 2021-03-08
Torch code in agents.torch and exploration_tools.torch

omardrwch:
This PR moves all torch agents to `rlberry.agents.torch` and RND to `rlberry.exploration_tools.torch`.

Two reasons for that:

* Installation of pytorch is optional, so it's good to isolate torch code;
* In the future, we'd like to include agents using other frameworks (e.g. jax, tensorflow), so it's better to organize the modules according to the frameworks.


xuedong (approved):
Perfecto.

PR #72, opened 2021-11-22
Improve doc

TimotheeMathieu:
In this PR I update the design of the doc using a slightly modified version of scikit-learn doc theme. 

I may need help to populate the User Guide, API and example sections of the doc, i.e. 
* I need someone to tell me what are the classes that should be integrated in the API menu, 
* I need more example with preferably a plot. Also I need to fix the stablebaselines example because for now it does not compile
* It would be nice to have a basic explanation in the user guide of some of the agents/environment. Basically, the user guide could use some of the content that @omardrwch will use to present RLBerry to his students I think.

**A little preview:** 
image
and the gallery of example that I added.
image


TimotheeMathieu:
Also I don't know how compatible readthedoc is with this new setup, if there is a problem, we can use github page instead (this is what is used by scikit-learn) which is more flexible I think.

TimotheeMathieu:
I finished setting up the preview of doc in PR, now there is a check in the PR that allows you to preview the documentation. And I made it work so I am sure that when I push, the documentation will compile.

For now, this doc is only a skeleton, there is a lot of content that have to be added bu I propose we push this PR and do separate PR for the content (in particular, I'd like to quickly make a page on how to contribute and the guidelines because there will soon be other contributors to rlberry from the INRIA team).

mmcenta:
Hello, I can review this if that's okay.

TimotheeMathieu:
Ok, no problem.

examples_old/Readme.md
```diff
@@ -0,0 +1,2 @@
+temporary directory for old examples.
+TODO : sort and include in doc if possible.
```
mmcenta, line 2: Maybe add an issue about this.

docs/themes/scikit-learn-fork/javascript.html
```diff
@@ -0,0 +1,149 @@
+{% if theme_google_analytics|tobool %}
```
mmcenta, line 1: I don't know enough JavaScript to review this properly, but I trust you!

mmcenta (approved):

mmcenta:
LGTM! We still need to sort out the examples, but we can merge the changes for now.

docs/themes/scikit-learn-fork/javascript.html
TimotheeMathieu: I don't know javascript either, I just copy/paste this from scikit-learn and anyway we don't have google analytics so it should not be a problem.

TimotheeMathieu:
Ok, I just added some classes in the API doc. I push to give the skeleton of the doc if people want to contribute, the content will follow in other PR.

Issue #70, opened 2021-11-18
Doc improvement proposal


TimotheeMathieu:
Proposition of improvements:

- [x] Add the link to https://rlberry.readthedocs.io in the "about" section in main github page 
- [x] Have an API documentation in the doc (which automatically copy the docstrings of major functions)
- [x] Support mathjax in the doc to use equations
- [x] Add some examples and visual illustrations and plots of the main functions (e.g. comparison of agents...)
- [x] Put most of the content of the Readme in the docs : I find the Readme to be too long and it tries to do the job of the doc. More generally, make a policy of having all of the docs in the doc ?
- [x] Maybe enhance the design of the doc a bit, add the logo... We could also inspire ourselves (or use) scikit-learn theme (this is also a sphinx theme) ? 
- [ ] Inspire ourselves of scikit-learn [contribution page](https://scikit-learn.org/dev/developers/contributing.html) for the contribution section of the doc
- [x] Have the doc constructed at PR time (using https://docs.readthedocs.io/en/stable/pull-requests.html) this will be useful to check that the doc is working typically are the code for the examples rendering a plot or not ?

I open the issue for brainstorming purposes and because I had these ideas when I saw the github for rlberry, if someone has other ideas, they can add them here and I may also want to add some more ideas later on. When we agree on some main ideas, I will make

mmcenta:
Hi,

So PR #72 solved most tasks on this issue except the contribution page, right?

TimotheeMathieu:
Yes, I plan to do a PR on the contribution page very soon to provide a guideline for new contributors, I know that some of the team members wanted to contribute but they were not sure how/where to do that. 

In particular, I will include some guidelines for making a new agent, maybe explain a bit git, the doc, what we want as docstrings... 

The guideline is not fixed yet, I will propose something (heavily influenced by scikit-learn guideline) and I will await comments on the PR so that we all agree on how to do things. For example, I am not sure whether we should encourage people to do fork and then PR into the main repo or if we just use different dev branches in the main repo and don't use forking. I personally prefer the fork solution but it will be up to discussion.

PR #77, opened 2021-11-29
[WIP] feat (doc): Doc contributing

TimotheeMathieu:
In this PR, I add a contribution page to the doc and I fix some broken links and delete some unneeded files in the doc. 
Should solve issue #70.

Feel free to react to the contribution guidelines, this is the place where we say what contributors must do. In particular I propose:

* Using forking and PR to contribute in order to avoid multiple branches on main repo and huge repo with long cloning time. Like this everyone can code on its own on his own forked repo.
* Using black to format the code and flake8 to check for  PEP8 violations (TODO : make an issue of adding this as checks at PR time)
* for docstring, minimal requirements should be = short description + parameters + return section. 
* All the tests must be green

I am also wondering whether we should put this rule : for PR that have a non-negligible effect on the library there should be at least one review. The problem is what is the definition of PR with non-negligible effect on the library ? Typically, I think that maintenance PR may be pushed without being reviewed but the line between negligible and non-negligible is a bit blurry in my mind.


TimotheeMathieu:
I merge for now even though it is only a draft. We will make corrections later if needed.

PR #78, opened 2021-11-30
Fix AgentHandler bug

omardrwch:
The `AgentHandler` class is used by `AgentManager` in order to save/load agents when returning to the main process after training agents in different processes/threads.

It calls `Agent.save()`, which uses pickle by default. This was causing bugs when the `env` and `eval_env` attributes of the agent were not pickleable. In this PR, both attributes are set to None before saving, and are restored by `Agent.load()` with the constructors provided by `AgentManager`.


TimotheeMathieu:
Maybe just do an example or tutorial or user guide entry in the doc to explain that, with a link from the docstring of AgentManager saying something like  "if it is not pickleable, see the example in the documentation". 

This example would be an example for user-defined save/load which can also be interesting independently.

PR #79, opened 2021-11-30
[MRG] feat(doc) : Render videos in doc

TimotheeMathieu:
In this PR I add a function to get the video in the rendering module and I use it to display videos in the documentation using an example of gridworld and acrobot. I think this is a good thing to have for a RL library.

PRO: Visually nice
CON: It takes a long time to compile which is not nice for readthedocs and maybe if we continue to add such examples we will not be able to construct the doc anymore (readthedoc is free and we have a time limit on runs).

Maybe the solution would be to plot only very simple environments or to construct the documentation offline and uploading the mp4 videos using github lfs for instance ? Or maybe there are some things I can optimize to reduce training time as well as conversion time (the time for ffmpeg to render the gif is rather long).
If someone has a better idea, please share it.


TimotheeMathieu:
Update : acrobot is too long to run in readthedocs. I guess we will have to generate the video offline, I will look into it.

TimotheeMathieu:
Update : now it works. I tried to use lfs because it would be cleaner but readthedocs does not support lfs.

TimotheeMathieu:
I simplified the process I included some doc on how to do this:

* Have an example python file formatted using the template given in the doc
* use `make video` to generate video and thumbnail
* push all of this on github

and you are good to go. 
Because I use this makefile I think this should work on any OS.

PR #88, opened 2021-12-21
[WIP] feat(test) Enhence checks (tests, rtd and coverage)

TimotheeMathieu:
In this PR, 
- I re-add the rendering tests that I removed before thanks to pytest-xvfb that allows one to test gui rendering without having to have a display. 
- I do a coverage report in azure-pipeline 

Questions:
- Are the coverage report best done in azure pipeline or in dedicated workflow (as it is done until now) ?
- Do we keep both test for linux with python 3.7 and 3.8 ? Do we add tests for python 3.7 for other OS or do we only support python 3.8 ?


TimotheeMathieu:
@omardrwch I have set up the codecov via azure pipeline. I think it is better because we get a report after doing a new commit in the PR (and not only for the first commit) and we can get more information by clicking on "details" if we need to. and I prefer to have them in the checks and not as a comment. What do you think ?

My proposition would be to remove the workflow that was previously used. The only thing is that for now, I cannot test the jax agents in azure pipeline, maybe we should try to test them later but they seem a bit instable so for now I excluded them from the test.

omardrwch:
@TimotheeMathieu Yes, I agree that it's better to have the codecov results in the checks instead of a comment!

And I think it's ok to ignore the jax tests for now, we have a single jax agent at the moment and it's mostly experimental.

Regarding the Python version, I think we can support only 3.8. Is there any reason to keep testing with 3.7?

TimotheeMathieu:
There is no solid reason to keep compatibility with 3.7 except that some people may prefer 3.7 because of some other library that is not compatible with 3.8... but let us keep things simple and support only 3.8. It will also remove one of the tests which is good.

TimotheeMathieu:
We have at most 7min check, which is acceptable I think because at least half of this is due to the installation of dependencies. Maybe later we can make an issue of having faster test & maybe try to see if all the dependencies are necessary but I think for now we stay like this.
I merge.

PR #90, opened 2022-01-08
[MRG] feat (plot) plot_writer_data enhencement

TimotheeMathieu:
In this PR I 
* change the parameter `fignum` to a matplotlib axis `ax` in `evaluate_agents` and `plot_writer_data`
* Add a parameter to `plot_writer_data` that record the plot in a file (via matplotlib's `savefig`).
* Add a `read_writer_data` function that returns the data to be plotted (can be used independently from plot_writer_data to have customized plots)
* Add the feature to read the last recorded pickle files and read the writer data.
* fix a few things for easier plot handling

The feature of reading from the pickle files is still experimental, in particular it depends heavily on the timestamps and the fact that the user has a compatible save/load function for his agent.


rlberry/manager/evaluation.py
```diff
@@ -84,14 +89,114 @@ def evaluate_agents(agent_manager_list,
 
     return output
 
+def read_writer_data(agent_manager,
```
omardrwch, line 92: I think that if `input_dir` is given, we don't actually need `agent_manager` as input. We could have a single argument for both, that is either:
* An instance of `AgentManager`
* A list of instances of `AgentManager`
* A directory (which contains all the info that we need to read the data)

TimotheeMathieu: Ok, good idea.

TimotheeMathieu:
I added a small explanation of thez fact that we don't have to refit if we want to replot in the plot_writer_data.py example. Both plot_writer_data and read_writer_data now take agent managers as input and load the files in which the agent managers are meant to be saved (this can be in several folders as long as it is specified in agent.output_dir).

TimotheeMathieu:
Ah, I noticed that when you input sns_kwargs, it will bug. I will fix it.

TimotheeMathieu:
all good no. Good to go.

omardrwch (approved):

PR #91, opened 2022-01-11
fix(agents): initialization issue of UCBVI

menardprr:
Fix the initialization of the value and bonus of UCBVI to v_max[hh] instead of v_max[0] in the stage independent case.

### DIFF
--- a/rlberry/agents/ucbvi/ucbvi.py
+++ b/rlberry/agents/ucbvi/ucbvi.py
@@ -3,7 +3,7 @@ import numpy as np
 
 import gym.spaces as spaces
 from rlberry.agents import AgentWithSimplePolicy
-from rlberry.agents.ucbvi.utils import update_value_and_get_action, update_value_and_get_action_sd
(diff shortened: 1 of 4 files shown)


omardrwch (approved):

rlberry/agents/ucbvi/ucbvi.py
```diff
@@ -3,7 +3,7 @@
 
 import gym.spaces as spaces
 from rlberry.agents import AgentWithSimplePolicy
-from rlberry.agents.ucbvi.utils import update_value_and_get_action, update_value_and_get_action_sd
+from rlberry.agents.ucbvi.utils import update_value_and_get_action, update_value_and_get_action_sd, backward_induction_reward_sd
```
omardrwch, line 6: The function `backward_induction_reward_sd` must be imported from `rlberry.agents.dynprog.utils`, no? This should fix the tests.

omardrwch (changes requested):
I think it is ok after fixing the import error

omardrwch (approved):

PR #97, opened 2022-01-17
[MRG] feat (agent and env) Bandits

TimotheeMathieu:
This is a module for Bandits


rlberry/agents/bandits/index_agents.py
```diff
@@ -0,0 +1,152 @@
...
+        self.index_function = index_function
+        self.recursive_index_function = recursive_index_function
+        self.record_action = record_action
+        self.phase = phase
+    def fit(self, budget=None, **kwargs):
+        n_episodes = budget
+        rewards = np.zeros(n_episodes)
+        actions = np.ones(n_episodes)*np.nan
+        # Initialization : pull all the arms once
+        for a in range(self.n_arms):
```
omardrwch, line 51: The `fit` method can be called several times, and we'd like to run this loop only the first time, right? So we could keep a total time counter and do something like:

```python
for t in range(n_episodes):
   self.total_time += 1  # initialized to 0 in __init__
   if self.total_time < self.n_arms:
       action = self.global_time
   else:
       ...  # indexes etc

```

TimotheeMathieu: Yes, good idea. Most of the time, I think in a bandit framework we don't need to continue a fit but I agree it could be useful, I will make the change.

TimotheeMathieu:
@AleShi94 I added an example with some subplot and custom plots. Tell me what you think !

PR #105, opened 2022-01-21
[WIP] Improve docs

omardrwch:
* Improved docs for:
      * `Agent`
      * `AgentManager`
* Unify `User Guide` and `Tutorials` sections.


rlberry/agents/agent.py
```diff
@@ -20,6 +20,10 @@
 class Agent(ABC):
     """Basic interface for agents.
 
+    .. note::
+
+        Classes that implement this interface should send ``**kwargs`` to :code:`Agent.__init__()`
+
```
TimotheeMathieu, line 26: proposition : instead of this, put at the end of the Parameters the lines 
```
...
    Parameters
    ----------
    ...
    _default_writer_kwargs : dict (optional)
        Parameters to initialize DefaultWriter (attribute self.writer).
    kwargs : key, value mappings
        keyword arguments passed down to :class:`~rlberry.agents.Agent`.__init__.
```
This will give a link to the Agent API page in the doc

TimotheeMathieu (commented):
This looks great ! 
Some minor comments.

rlberry/agents/agent.py
```diff
@@ -276,49 +291,13 @@ def load(cls, filename, **kwargs):
...
-        If true, makes a deep copy of the environment.
-    seeder : rlberry.seeding.Seeder, int, or None
-        Object for random number generation.
-    _execution_metadata : ExecutionMetadata (optional)
-        Extra information about agent execution (e.g. about which is the process id where the agent is running).
-    _default_writer_kwargs : dict (optional)
-        Parameters to initialize DefaultWriter (attribute self.writer).
-    .. note::
-        Classes that implement this interface should send ``**kwargs`` to :code:`Agent.__init__()`
-
```
TimotheeMathieu, line 303: I agree that this is a copy/paste of Agent, hence redundant, but I would still keep it because people don't necessarily think of going to the Agent doc to have a doc on AgentWithSimplePolicy. Moreover, if you want this docstring to be usable in python with `?AgentWithSimplePolicy`, it would be annoying to have to import Agent to do `?Agent` to know the doc on AgentWithSimplePolicy. The best solution would be to automatically inherit in the doc but I don't know if this is doable. probably not.

docs/user_guide.rst
```diff
@@ -6,40 +6,28 @@
...
-
-.. toctree::
-
-  basics/evaluate_agent.rst
-  basics/compare_agents.rst
-  basics/seeding.rst
-
-
-How to contribute ?
+Contributing
```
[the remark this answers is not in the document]
omardrwch: Ok, for now I'll change it back to how it was!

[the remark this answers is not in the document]
TimotheeMathieu: After some research : this is a inherent behavior of rst.This can't really be changed. However we can target contributing doc and not the title using the `doc` keyword : 
```


Contributing
============

If you want to contribute to rlberry, check out :doc:`the contribution guidelines<contributing>`.
```
This would work, but on the other hand, all the :ref: links to contributing will send to the User Guide, we would have to change ref to doc in all the links to contributing (meaning the link in the User guide and the link in index.rst)

rlberry/agents/agent.py
omardrwch: I think it would be weird to have it in the parameters section, since the first thing we do in `Agent` is an assertion

```python
 assert kwargs == {}, "Unknown parameters sent to agent:" + str(kwargs.keys())
```

The ``**kwargs`` are here just to ensure that `AgentManager` can send parameters to `Agent` that are not handled by the user (such as output_dir, thread/process id etc.), in case we add more parameters of this kind to the `Agent` constructor.

TimotheeMathieu: Ok. But at the same time, from a user perspective this is a parameter and the user see it in the parameters of init, he will search for it in the parameter section. You suppose that the user will read your note before going to the parameters...which is not certain I think. 

For reference this is also how it is done in [seaborn](https://seaborn.pydata.org/generated/seaborn.lineplot.html)

TimotheeMathieu (approved):

PR #107, opened 2022-01-25
Sac

riccardodv:
I am creating a new branch where Alena and I can collaborate in an easier way on the implementation of SAC, I hope I am doing things properly but let me know.


rlberry/agents/torch/sac/sac.py
```diff
@@ -0,0 +1,307 @@
...
+        Horizon.
+    gamma : double
+        Discount factor in [0, 1].
+    entr_coef : double
+        Entropy coefficient.
+    learning_rate : double
+        Learning rate.
+    optimizer_type: str
+        Type of optimizer. 'ADAM' by defaut.
+    k_epochs : int
```
TimotheeMathieu, line 44: Isn't this what is called horizon in other agent ? I have some troubles understanding which is which between horizon, k_epochs and batch_size. Maybe you could explain more what is horizon ?

rlberry/agents/torch/sac/sac.py
```diff
@@ -0,0 +1,307 @@
...
+        Arguments for the UncertaintyEstimatorWrapper
+    device : str
+        Device to put the tensors on
+
+    References
+    ----------
+    Mnih, V., Badia, A.P., Mirza, M., Graves, A., Lillicrap, T., Harley, T.,
+    Silver, D. & Kavukcuoglu, K. (2016).
+    "Asynchronous methods for deep reinforcement learning."
+    In International Conference on Machine Learning (pp. 1928-1937).
```
TimotheeMathieu, line 69: Maybe you can add 
```
See also :class:`rlberry.agents.torch.A2CAgent` 
```
this will make a link in the doc to A2CAgent, because the agents are related.

TimotheeMathieu (commented):
Thanks, LGTM. I am not very familar with this algo and with deep agents in general so I may not be very helpful. 
For the branch, it is ok to do it like you did. 

- You have to add some tests (pytest tests) to ensure that the algo works (this is linked to the coverage checks which is failing for now). The test must be fast to compile so often you will only use 1 iteration or something like that, just to check that you can.
- It would be good to add a working example in the doc. It is good practice to always have an example. You can also just add this agent in an example in which we compare several agents.
- for the "format" check that is currently failing, you have to black everything. i.e. this should work:
```
pip install black
black rlberry/agents/torch/sac/
```

Thanks

rlberry/agents/torch/sac/sac.py
riccardodv: ah yes this stuff is not finished and working properly. It was just the beginning. should we commit here just when things work properly?

TimotheeMathieu: No you can commit before it is ready, no worries. This way we can help if there are problems specific to this PR, you can ask questions about it here.

PR #110, opened 2022-02-01
feat (tests) add test of compatibility for agents and environments.

TimotheeMathieu:
The goal is to add check functions and tests to assess compatibility of agents and envs in rlberry.

- [ ] test seeding in discrete state agents
- [ ] test seeding in continuous state agents
- [ ] test seeding in environments
- [ ] test compatibility with agent manager
- [ ] test compatibility with hyperparameter optimization
- [ ] test compatibility with save and load functions (in particular, must have _writer for plots)

### DIFF
--- a/rlberry/tests_compatible/tests_agent.py
+++ b/rlberry/tests_compatible/tests_agent.py
@@ -1,0 +1,34 @@
+import pytest
+from rlberry.agents import *
+from rlberry.agents.torch import *
+from rlberry.envs.finite import Chain
+from rlberry.utils.check_discrete_action_agent import (
+    check_finiteMDP_agent,
+    check_continuous_state_agent,
+)
+
+FINITE_MDP_AGENTS = [
+    ValueIterationAgent,
+    MBQVIAgent,
+    UCBVIAgent,
+]
+
+
+CONTINUOUS_STATE_AGENTS = [
+    RSUCBVIAgent,
+    RSKernelUCBVIAgent,
+    OptQLAgent,
+    LSVIUCBAgent,
+    DQNAgent,
+    REINFORCEAgent,
+]
+
+
+@pytest.mark.parametrize("Agent", FINITE_MDP_AGENTS)
+def test_finite_state_agent(Agent):
+    assert check_finiteMDP_agent(Agent)
+
+
+# @pytest.mark.parametrize("Agent", CONTINUOUS_STATE_AGENTS)
+# def test_continuous_state_agent(Agent):
+#     assert check_continuous_state_agent(Agent)
--- a/rlberry/tests_compatible/tests_envs.py
+++ b/rlberry/tests_compatible/tests_envs.py
@@ -1,0 +1,25 @@
+from rlberry.utils.check_env import check_env
+from rlberry.envs import Acrobot
+from rlberry.envs.benchmarks.ball_exploration import PBall2D
+from rlberry.envs.benchmarks.generalization.twinrooms import TwinRooms
+from rlberry.envs.benchmarks.grid_exploration.apple_gold import AppleGold
+from rlberry.envs.benchmarks.grid_exploration.nroom import NRoom
+from rlberry.envs.classic_control import MountainCar
+from rlberry.envs.finite import Chain, GridWorld
+import pytest
+
+ALL_ENVS = [
+    Acrobot,
+    PBall2D,
+    TwinRooms,
+    AppleGold,
+    NRoom,
+    MountainCar,
+    Chain,
+    GridWorld,
+]
+
+
+@pytest.mark.parametrize("Env", ALL_ENVS)
+def test_env(Env):
+    check_env(Env())
--- a/rlberry/utils/__init__.py
+++ b/rlberry/utils/__init__.py
@@ -1,0 +1,1 @@
+from .check_discrete_action_agent import check_finiteMDP_agent
--- a/rlberry/utils/check_discrete_action_agent.py
+++ b/rlberry/utils/check_discrete_action_agent.py
@@ -1,0 +1,58 @@
+from rlberry.envs import Chain
+from rlberry.envs.benchmarks.ball_exploration import PBall2D
+from rlberry.agents import ValueIterationAgent
+from rlberry.manager import AgentManager, evaluate_agents
+import numpy as np
+
+
+def check_finiteMDP_agent(Agent):
+    env_ctor = Chain
+    env_kwargs = {}
+
+    agent1 = AgentManager(
+        Agent, (env_ctor, env_kwargs), fit_budget=10, n_fit=1, seed=42
+    )
+    agent2 = AgentManager(
+        Agent, (env_ctor, env_kwargs), fit_budget=10, n_fit=1, seed=42
+    )
+
+    agent1.fit()
+    agent2.fit()
+    env = env_ctor(**env_kwargs)
+    state = env.reset()
+    action1 = agent1.agent_handlers[0].policy(state)
+    action2 = agent2.agent_handlers[0].policy(state)
+
+    return action1 == action2
+
+
+def check_continuous_state_agent(Agent):
+    env_ctor = PBall2D
+    env_kwargs = {}
+
+    agent1 = AgentManager(
+        Agent,
+        (env_ctor, env_kwargs),
+        fit_budget=2,
+        n_fit=1,
+        init_kwargs={"horizon": 3},
+        seed=42,
+    )
+    agent2 = AgentManager(
+        Agent,
+        (env_ctor, env_kwargs),
+        fit_budget=2,
+        n_fit=1,
+        init_kwargs={"horizon": 3},
+        seed=42,
+    )
+
+    agent1.fit()
+    agent2.fit()
+    env = env_ctor(**env_kwargs)
+    state = env.reset()
+
+    action1 = agent1.agent_handlers[0].policy(state)
+    action2 = agent2.agent_handlers[0].policy(state)
+
+    return np.mean(action1 == action2) == 1
--- a/rlberry/utils/check_env.py
+++ b/rlberry/utils/check_env.py
@@ -1,0 +1,23 @@
+from rlberry.seeding import safe_reseed
+from rlberry.seeding import Seeder
+from copy import deepcopy
+import numpy as np
+
+seeder = Seeder(42)
+
+
+def check_env(env):
+    # Test if env adheres to Gym API
+    action = env.action_space.sample()
+    safe_reseed(env, Seeder(42))
+    env.reset()
+    a = env.step(action)[0]
+
+    safe_reseed(env, Seeder(42))
+    env.reset()
+    b = env.step(action)[0]
+    if hasattr(a, "__len__"):
+        print(a, b)
+        assert np.mean(np.array(a) == np.array(b)) == 1
+    else:
+        assert a == b


TimotheeMathieu:
The codecov patch does not pass because I don't test the test files from gym. I actually don't know what is the best policy concerning including or not the test files in the coverage report, maybe we could remove them from the coverage report.

rlberry/agents/__init__.py
```diff
@@ -10,5 +10,6 @@
 from .mbqvi import *
 from .optql import *
 from .ucbvi import *
-from .rlsvi import *
-from .psrl import *
+from .linear import *
+from .psrl import PSRLAgent
```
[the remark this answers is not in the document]
TimotheeMathieu: Yes, and there are star import in practically all the __init__, maybe a new PR that rewrite all star import as non-star import would be good. Added to todo.

rlberry/tests/test_agent.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class OneHotLSVI(LSVIUCBAgent):
+    def __init__(self, env, **kwargs):
+        def feature_map_fn(_env):
+            return OneHotFeatureMap(5, 2)  # values for Chain
+
+        LSVIUCBAgent.__init__(
+            self, env, feature_map_fn=feature_map_fn, horizon=10, **kwargs
+        )
```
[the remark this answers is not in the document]
omardrwch: I agree that it might be a good idea to always provide default kwargs for agents, which would avoid wrapping here (but I think it's something for another PR).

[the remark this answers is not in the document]
TimotheeMathieu: I will keep it as is but maybe in a futur PR we may want to set a guideline like "all parameters that are not env should have a default" ?

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+from rlberry.manager import AgentManager
+import numpy as np
+from rlberry.seeding import set_external_seed
+import tempfile
+import os
+
+SEED = 42
+
+
+def check_agent_manager(Agent, continuous_state=False):
```
omardrwch, line 12: I see that the flag `continuous_state` is only used to get a constructor for the environment. I think it's better to replace it by a tuple `(env_constructor, env_kwargs)`, so that the users can test their own combinations of agent+environment.

I'm saying this because there are many observation spaces that aren't continuous or discrete. We can have `Dict` spaces with possibly nested continuous/discrete observations, and currently the `check_agent_manager` method (and others) assume either a `Discrete` or `Box` observation space.

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+        environment state
+    agent1: rlberry agent
+        first agent to be compared
+    agent2: rlberry agent
+        second agent to be compared
+    n_checks : int, default=5
+        number of checks to do.
+    """
+    result = True
+    set_external_seed(SEED)
```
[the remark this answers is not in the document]
TimotheeMathieu: I actually tried safe_reseed and for some reason it does not work with PPOAgent, AVECPPOAgent and REINFORCEAgent. on the other hand, there is no problem with what I did if we suppose that the agents have the same seed (if they don't have they are not almost_equal).

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    actions1 = []
+    for f in range(5):
+        # test on 5 actions
+        actions1.append(agent1.agent_handlers[0].policy(state))
+    set_external_seed(SEED)
+    for f in range(5):
+        # test on 5 actions
+        action2 = agent2.agent_handlers[0].policy(state)
+        if np.mean(actions1[f] == action2) != 1:
```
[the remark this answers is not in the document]
TimotheeMathieu: well, it does the same thing and I don't really have a preference it is just that I thought of this on the moment. Do you prefer no.all ?

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    result = True
+    set_external_seed(SEED)
+    actions1 = []
+    for f in range(5):
+        # test on 5 actions
+        actions1.append(agent1.agent_handlers[0].policy(state))
+    set_external_seed(SEED)
+    for f in range(5):
+        # test on 5 actions
+        action2 = agent2.agent_handlers[0].policy(state)
```
omardrwch, line 68: Here, we are assuming that the agents have a `policy` method, which is not the case for all agents. For instance, agents for POMDPs will not have a policy that is a function of an observation only, since they must take the past (or a hidden state) into account.

Suggested solution: instead of comparing the actions, compare the output of the `eval` method of the agents.

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    ----------
+
+    Agent: rlberry agent module
+        Agent class that we want to test.
+
+    continuous_state: bool, default=False
+        whether we test on a discrete state environment or a continuous one.
+    """
+    agent1 = check_agent_manager(Agent, continuous_state)
+    agent2 = check_agent_manager(Agent, continuous_state)
```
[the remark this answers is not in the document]
TimotheeMathieu: because it is faster than rewriting the fit and check_agent_manager already fit so I figured, let us use it to fit an agent.

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    state = env.reset()
+    result = check_agent_almost_equal(state, agent1, agent2)
+
+    assert result, "Agent not reproducible (different seed give different results)"
+
+
+def check_rl_agent(Agent, continuous_state=False):
+    """
+    Check agent manager compatibility  and check reproducibility/seeding.
+    Will raise an exception if there is something wrong with the agent.
```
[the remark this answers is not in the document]
TimotheeMathieu: There are already exception in each individual check and I don't want to wrap all of this in a `check_rl_agent` exception because this would hide the real error/exception.

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    if continuous_state:
+        env_ctor = PBall2D
+    else:
+        env_ctor = Chain
+    env_kwargs = {}
+    try:
+        agent = AgentManager(
+            Agent, (env_ctor, env_kwargs), fit_budget=5, n_fit=1, seed=SEED
+        )
+        agent.fit()
```
omardrwch, line 34: Suggestion: also test eval

```python
agent.fit()
agent.eval_agents()
```

rlberry/utils/check_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    Check that two agents are almost equal in the sense that they give the same
+    actions n_checks times on a given state.
+
+
+    Parameters
+    ----------
+
+    state: state
+        environment state
+    agent1: rlberry agent
```
omardrwch, line 52: Agent or agent manager? It seems that you are using a manager below.

rlberry/utils/check_gym_env.py
```diff
@@ -0,0 +1,340 @@
...
+    seed_param = signature.parameters.get("seed")
+    # Check the default value is None
+    if seed_param is not None and seed_param.default is not None:
+        logger.warn(
+            "The default seed argument in reset should be `None`, "
+            "otherwise the environment will by default always be deterministic"
+        )
+
+
+def check_gym_env(
```
omardrwch, line 285: What prevents us from just importing this method from gym instead of copying the code for the tests? (also, we are not using the same docstring style as gym)

omardrwch (changes requested):
Main suggestions:

* Send `(env_constructor, env_kwargs)` instead of `continous_state` to the check functions.

* When checking if two agents are equal, use `eval` method (implemented by all agents) instead of `policy` (only available for `AgentWithSimplePolicy`.

Please let me know what you think!

TimotheeMathieu:
For the codecov issue, I agree with you and ideally check_gym_env should be tested and I added gym's test but their test is very minimalist and it would be very tedious to check all of the error messages from check_gym_env.

rlberry/tests/test_agent.py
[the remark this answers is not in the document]
omardrwch: Ok! Yes, I think that it'd be a nice guideline! But only for the agents that we add/test in the library (so we don't add any constraint to user-only code).

rlberry/utils/check_gym_env.py
TimotheeMathieu: I removed some of the tests from gym that were failing on our environments. I don't remember which ones but it was mostly due to some conventions and I beleived it is not necessary to change all our envs to agree with their convention. Do you agree or should we change our environments ?

rlberry/utils/check_agent.py
TimotheeMathieu: Ok but at the same time my check aimed at being simple and if I add what you suggest then the user can't just `check_rl_agent(My_awesome_agent)`, he has to also define an environment. And also I prefered to do this because I am sure that Chain and PBall2D are fine, if the user has a problem he will not know whether this is due to the agent or the environment. 
But at the same time you are right that more generality would be nice. Maybe I can have the two possible parameters : if env is given, use it if not use default env.

rlberry/utils/check_agent.py
TimotheeMathieu: Ok, but at the same time there can be some problem with eval, for instance the user coded his agent with eval always returning 0, this would work. Or for bandits agent in particular, eval is just compare the optimal mean but two bandit agents can be different but still have the same optimal action (optimal arm identified). 
Is there a way to check without using eval ?

rlberry/utils/check_agent.py
TimotheeMathieu: same reservation as above.

rlberry/utils/check_agent.py
TimotheeMathieu: You are right, agent manager.

rlberry/utils/check_gym_env.py
TimotheeMathieu: Mainly if I remember right this was some typing issues. Like an state returned by env.step() must be int, and things like that. Some of ours are float. And also some problem because of the differences between gym.spaces and rlberry.spaces.

rlberry/utils/check_agent.py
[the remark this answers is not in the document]
mmcenta: I think `np.all` more clearly conveys what you're trying to do. Also, I'm always afraid of doing float comparisons.

rlberry/utils/check_agent.py
omardrwch: So we can have different checks according to the type of the agent:
* If `AgentWithSimplePolicy`, we keep the test as it is; 
* If not (only `Agent` is implemented), then we use `eval`.

In general (for RL), the output of `eval` will depend on the policy implemented/followed by the agent, so it should be good in most cases.

rlberry/utils/check_agent.py
omardrwch: It's just to check that `eval` is also running. If I remember correctly, a call to `fit` will only store the agent instances in disk (`dump` method of `AgentHandler`), whereas we need to load the agent from the disk (`AgentHandler.load`) to make `eval_agents` work. Sometimes `dump` works, but not `load`.

rlberry/utils/check_agent.py
TimotheeMathieu: Or we can have a parameter that say `compare_with="eval"` or `compare_with="policy"` or any string that correspond to the name of a method that returns a real number, would that be alright ?

The solution with `AgentWithSimplePolicy` vs `Agent` is also alright but the solution with compare_with would allow maybe more generalization.

rlberry/utils/check_agent.py
TimotheeMathieu: `load` is now checked in `check_save` (line 184 of `check_agent.py`). Maybe I can change the name to `check_save_load`

TimotheeMathieu: but yeah, I can add it here, there is no such thing as too much testing :)

rlberry/utils/check_agent.py
omardrwch: Yes, I think that `compare_with` is a nice solution, since it doesn't hide from the user the criterion that we use to compare the agents!

rlberry/utils/check_agent.py
omardrwch: Alright! But then there is some redundancy if we keep both `env` and `continuous_state` as parameters. What do you think of using a single parameter `env` that can be either:
* a string `continuous_state`
* a string `discrete_state`
* a tuple (constructor, kwargs)
?

It would also allow us to add extra options later, as `continuous_state_action` etc.

rlberry/utils/check_agent.py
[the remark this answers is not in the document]
TimotheeMathieu: For safe_reseed I agree and also I need the set_external_seed even though the seed has already been specified in AgentManager, this is due to torch using its own seed. Maybe we could do something to merge all the seeds ? I don't know, @omardrwch what do you think ?

For agents with different RNG, yes but I don't know how to test that two agents are the same because it depends on how they are describe (weights for neural networks, nulber of pulls for bandits...). For now I will keep it like that because in fact I use the test only with AgentManagers with the same seed and I don't need it for any other application but that may change in the futur.

rlberry/utils/check_agent.py
TimotheeMathieu: Ah, there is a problem with eval_agent that takes too much time. For now I will not use it.

rlberry/utils/check_agent.py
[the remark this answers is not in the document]
omardrwch: What do you mean by "merge all seeds"? In the library, we take the following approach:

* Each agent or environment has its own random number generator (rng) and, consequently, its own seed. This is very important to avoid global seeds, which are horrible for multiprocessing/threading (where we can't control how global variables are handled). Global seeds are also horrible in general, in the sense that if agents A and B rely on a global seed, running A then B is different from running B then A (which is precisely our issue with pytorch here).
* Not all libraries handle seeds that way, that is, they use global seeds (as pytorch, but not jax :D), and that is why we have the `set_external_seed` function. **This function is only supposed to be called one time in the beginning of each process** to avoid weird behaviors with respect to reproducibility and to make sure that all seeds used in a single process are independent.  By the way, since we're calling it  in `check_agent`, I believe this should raise a **big warning** to the user.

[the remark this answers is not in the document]
TimotheeMathieu: Ok, thanks this makes things clearer for me. Maybe I can include a message that say "check that reproducible if a global seed is given" in the docstring.

rlberry/utils/check_agent.py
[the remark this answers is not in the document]
mmcenta: I think I was surprised that a `check_`... function returns something because usually, they return nothing or a bool. However, I am having a hard time coming up with an alternative - maybe adding a `safe_manager_wrap` function that tries to wrap an agent (similarly to `safe_reseed`). Then. we could use it for the checks and also expose it so people can see.

rlberry/utils/check_agent.py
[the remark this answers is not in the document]
mmcenta: Oh, I meant this just as a rewrite of the "Will raise an exception if there is something wrong with the agent." part of the docstring since it seemed long. When I feel like a phrase in a docstring feels long or colloquial, I usually just google docstrings from other libraries (e.g. numpy, scikit) and copy their style. This is what I've done here!

rlberry/utils/check_agent.py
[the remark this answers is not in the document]
TimotheeMathieu: Ok, good idea, I will use _fit_agent_manager that returns a fitted agent and check_agent_manager that just do the check. You are right this is easier to understand.

rlberry/utils/check_gym_env.py
omardrwch: Ok, no problem, I agree to keep this adaptation you made!

rlberry/utils/check_agent.py
TimotheeMathieu: yes but the new code allow even more generality ! you can define a method like "agent.evaluation_with_weird_name" and use it in compare with, not only policy and eval !

TimotheeMathieu: I checked eval in the test of the check (`test_check.py` file)

TimotheeMathieu:
@mmcenta and @omardrwch , now it should be ok I think.

PR #113, opened 2022-02-04
feat (doc) Changelog

TimotheeMathieu:
Add a changelog to the docs.


TimotheeMathieu:
@omardrwch please check and add any change I missed.

docs/changelog.rst
```diff
@@ -0,0 +1,16 @@
+.. _changelog:
+
+Changelog
+=========
+
+Version 0.2.0
```
omardrwch, line 6: Should it be 0.3.0?

TimotheeMathieu: I don't know, is it current version or future one ? It is hard to know because you can't really plan in advance what will be the number of the futur version (i.e. we don't know if we won't include some drastic changes in the next version which will change the first number). Maybe a more easy to understand would be 

```
Unreleased 
---------------

blah blah

Version 0.2.1
-----------------

blah blah
```

omardrwch: I think it's a good idea. Instead of "unreleased" we could put "development" or "under development"

TimotheeMathieu: Something like that ok ?

PR #115, opened 2022-02-08
[WIP] Implementation of TD3 (experimental)

omardrwch:
Implementation of the TD3 algorithm.

Paper: Fujimoto et al. (2018) https://arxiv.org/abs/1802.09477
See also: https://spinningup.openai.com/en/latest/algorithms/td3.html

TODO:

* Improve docs of replay buffer (example docstring)


menardprr:
Why not implementing Ddpg before? Or is it possible to easily obtain Ddpg from the current implementation of Td3?

omardrwch:
> Why not implementing Ddpg before? Or is it possible to easily obtain Ddpg from the current implementation of Td3?

Because I already had something like TD3 coded somewhere. And yes, I think it shouldn't be hard to obtain DDPG from it!

xuedong:
Just for record regarding DDPG:

Paper: https://arxiv.org/abs/1509.02971
See also: https://spinningup.openai.com/en/latest/algorithms/ddpg.html

KohlerHECTOR:
So far, we think we dont need new agents. As such, this can be reopened when need but now, we clean the git.

PR #118, opened 2022-02-14
[MRG] feat (bandits) Thompson sampling 

TimotheeMathieu:
Add Thompson sampling Bandit algorithm with gaussian prior or beta prior. 
Create a base class for bandits algorithm with custom save & load functions (called BanditWithSimplePolicy)
Add examples and tests
Reorganize bandit doc and examples


TimotheeMathieu:
I merge. for now Bandits are still pretty experimental so I will not ask for a review but instead I merge so that people can test it more easily, let us review when it is more stable.

Issue #116, opened 2022-02-09
Agent manager is crashing when I try to parallelize a simple piece of code


riccardodv:
When I run the following code with AgentManager and parallelization 

```python
import numpy as np
from rlberry.manager import AgentManager
from rlberry.envs.benchmarks.ball_exploration import PBall2D
from rlberry.agents.torch.a2c.a2c import A2CAgent

env_ctor = PBall2D
env_kwargs = dict()
env = env_ctor(**env_kwargs)

agent = AgentManager(A2CAgent, (env_ctor, env_kwargs), fit_budget=100, n_fit=3, max_workers=2, parallelization="process")
agent.fit()
```

results in the following error

```
RuntimeError: 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

Constructing an agent with AgentManager and adding `mp_context = fork` seems to solve the problem but still it seems a problem.

TimotheeMathieu:
Did you try to execute this code in a notebook ? The default `mp_context = spawn` do not work in a notebook, the parallelization gets messed up and you have to use fork.

TimotheeMathieu:
To follow up, here is a working script :

```
import numpy as np
from rlberry.manager import AgentManager
from rlberry.envs.benchmarks.ball_exploration import PBall2D
from rlberry.agents.torch.a2c.a2c import A2CAgent

env_ctor = PBall2D
env_kwargs = dict()
env = env_ctor(**env_kwargs)

agent = AgentManager(A2CAgent, (env_ctor, env_kwargs), fit_budget=100, n_fit=3, max_workers=2, parallelization="process")
if __name__ == '__main__':
    agent.fit()
```
In short, the explanation is that if you use `paralellization="process"` and `mp_context=spawn` then the parallel computing code must be "protected" behing a `if __name__ == '__main__':`.

* `paralellization="process"`  is to have real paralellization, without this it is not really parallel except if you do a lot of numpy in each fit
* From what I understood, `mp_context=spawn` is mainly because `mp_context=fork` do not work with jax

TimotheeMathieu:
Solved by PR#120, I close.

PR #120, opened 2022-02-15
[MRG] feat (check) add explanation about spawn error

TimotheeMathieu:
Should solve Issue #116

When fitting an agent manager, if the code is not behind ` if __name__ == '__main__':`, throw the following error.

```
Warning: in AgentManager, if mp_context='spawn' and
                        parallelization="process" then the script must be run *
                        outside a notebook and protected by a  if __name__ == '__main__':
                                   For instance :

                                       agent = AgentManager(UCBVIAgent,(Chain, {}),
                                                            mp_context="spawn",
                                                            parallelization="process")
                                       if __name__ == '__main__':
                                           agent.fit(10)
```

### DIFF
--- a/rlberry/manager/agent_manager.py
+++ b/rlberry/manager/agent_manager.py
@@ -552,9 +552,9 @@ class AgentManager:
                 _check_not_importing_main()
             except RuntimeError as exc:
                 raise RuntimeError(
-                    """Warning: in AgentManager, if mp_context='spawn'
-                                   then the script must be run outside a notebook
-                                   and protected by a  if __name__ == '__main__':
+                    """Warning: in AgentManager, if mp_context='spawn' and
+                        parallelization="process" then the script must be run *
+                        outside a notebook and protected by a  if __name__ == '__main__':
                                    For instance :
 
                                        agent = AgentManager(UCBVIAgent,(Chain, {}),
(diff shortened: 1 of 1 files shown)


riccardodv (approved):

rlberry/manager/agent_manager.py
```diff
@@ -545,6 +546,25 @@ def fit(self, budget=None, **kwargs):
...
+                raise RuntimeError(
+                    """Warning: in AgentManager, if mp_context='spawn' and
+                        parallelization="process" then the script must be run *
+                        outside a notebook and protected by a  if __name__ == '__main__':
+                                   For instance :
+
+                                       agent = AgentManager(UCBVIAgent,(Chain, {}),
+                                                            mp_context="spawn",
+                                                            parallelization="process")
+                                       if __name__ == '__main__':
```
omardrwch, line 563: In the example, I think it's better to put everything inside the `if __name__ == '__main__'`:

```python
if __name__ == '__main__':
    agent = AgentManager(
                       UCBVIAgent,
                      (Chain, {}),
                      mp_context="spawn",
                      parallelization="process")
    agent.fit(10)
```
otherwise the `AgentManager` constructor will be re-run in each new interpreter created by spawn.

omardrwch (changes requested):

rlberry/manager/agent_manager.py
TimotheeMathieu: Good catch, I actually always used it with only the fit I did not know that there was much of a difference. Thanks.

omardrwch (approved):

PR #124, opened 2022-02-17
[MRG] feat (bandits) Randomized agents

sauxpa:
* Add RandomizedAgent class, inspired by IndexAgent, with default policy being EXP3 with standard, anytime hyperparam,
* Add AdversarialBandit environment, where all potential rewards are fixed in advance (oblivious opponent). 
* Refactor envs/bandits/bandits.py to remove redundant syntax when instantiating a child class.
* Refactor examples/demo_bandits to display pseudo-regret (best mean - mean of chosen arm) rather than regret (best mean - reward of chosen arm).
   * Pseudo-regret is the standard measure in bandit, and it is by definition nondecreasing, whereas regret as it is now is noisy.
* Update tests
   * Change utils/check_bandit_agent to test on BernoulliBandit rather than NormalBandit. 
      * Reason is some algo (EXP3) crucially rely on bounded rewards (by default [0, 1]). I think it's safer to test by default on Bernoulli rather than Gaussian. Also I explicitly pass the env in argument in the test for readability. 
   * Add test for BernoulliBandit env. 
   * Explicitly fixes the seed (add seed argument).

### DIFF
--- a/rlberry/agents/bandits/index_agents.py
+++ b/rlberry/agents/bandits/index_agents.py
@@ -36,7 +36,6 @@ class IndexAgent(BanditWithSimplePolicy):
 
     def __init__(self, env, index_function=None, **kwargs):
         BanditWithSimplePolicy.__init__(self, env, **kwargs)
-        self.n_arms = self.env.action_space.n
         if index_function is None:
 
             def index(r, t):
--- a/rlberry/agents/bandits/index_agents.py
+++ b/rlberry/agents/bandits/index_agents.py
@@ -38,6 +38,7 @@ class IndexAgent(BanditWithSimplePolicy):
         BanditWithSimplePolicy.__init__(self, env, **kwargs)
         self.n_arms = self.env.action_space.n
         if index_function is None:
+
             def index(r, t):
                 return np.mean(r) + np.sqrt(np.log(t**2) / (2 * len(r)))
 
--- a/rlberry/agents/bandits/index_agents.py
+++ b/rlberry/agents/bandits/index_agents.py
@@ -34,11 +34,10 @@ class IndexAgent(BanditWithSimplePolicy):
 
     name = "IndexAgent"
 
-    def __init__(self, env, index_function=lambda rew, t: np.mean(rew), **kwargs):
+    def __init__(self, env, index_function=None, **kwargs):
         BanditWithSimplePolicy.__init__(self, env, **kwargs)
         self.n_arms = self.env.action_space.n
         if index_function is None:
-
             def index(r, t):
                 return np.mean(r) + np.sqrt(np.log(t**2) / (2 * len(r)))
 
@@ -52,25 +51,25 @@ class IndexAgent(BanditWithSimplePolicy):
         rewards = np.zeros(horizon)
         actions = np.ones(horizon) * np.nan
 
-        indexes = np.inf * np.ones(self.n_arms)
+        indices = np.inf * np.ones(self.n_arms)
         for ep in range(horizon):
             if self.total_time < self.n_arms:
                 action = self.total_time
             else:
-                indexes = self.get_indexes(rewards, actions, ep)
-                action = np.argmax(indexes)
+                indices = self.get_indices(rewards, actions, ep)
+                action = np.argmax(indices)
             self.total_time += 1
             _, reward, _, _ = self.env.step(action)
             rewards[ep] = reward
             actions[ep] = action
 
-        self.optimal_action = np.argmax(indexes)
+        self.optimal_action = np.argmax(indices)
         info = {"episode_reward": np.sum(rewards)}
         return info
 
-    def get_indexes(self, rewards, actions, ep):
+    def get_indices(self, rewards, actions, ep):
         """
-        Return the indexes of each arms.
+        Return the indices of each arm.
 
         Parameters
         ----------
@@ -85,16 +84,16 @@ class IndexAgent(BanditWithSimplePolicy):
             current iteration/epoch
 
         """
-        indexes = np.zeros(self.n_arms)
-        for a in range(self.n_arms):
-            indexes[a] = self.index_function(rewards[actions == a], ep)
-        return indexes
+        indices = np.zeros(self.n_arms)
+        for a in self.arms:
+            indices[a] = self.index_function(rewards[actions == a], ep)
+        return indices
 
 
 class RecursiveIndexAgent(BanditWithSimplePolicy):
     """
     Agent for bandit environment using Index-based policy like UCB with
-    recursive indexes.
+    recursive indices.
 
     Parameters
     -----------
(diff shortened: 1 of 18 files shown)


sauxpa:
Windows build now fails but seems to be unrelated to my commit (used to pass on previous commits, now fails after reverting).

rlberry/manager/evaluation.py
```diff
@@ -304,6 +304,9 @@ def plot_writer_data(
     if ax is None:
         figure, ax = plt.subplots(1, 1)
 
+    # PS: in the next release of seaborn, ci should be deprecated and replaced
+    # with errorbar, which allows to specifies other types of confidence bars,
+    # in particular quantiles.
```
TimotheeMathieu, line 309: Nice ! we may want to change to quantile then. As of now, we chose the 'sd' ci because choosing the permutation ci (ie argument ci = 75 for instance) is costly with large dataset.

rlberry/utils/check_bandit_agent.py
```diff
@@ -1,8 +1,8 @@
-from rlberry.envs.bandits import NormalBandit
+from rlberry.envs.bandits import BernoulliBandit
 from rlberry.manager import AgentManager
 
 
-def check_bandit_agent(Agent, environment=NormalBandit):
+def check_bandit_agent(Agent, environment=BernoulliBandit, seed=None):
```
TimotheeMathieu, line 5: the seed has to be fixed. It can be any int, but it can't be None because if it is None, there would be reproducibility issues. Maybe you can change the default to some int (42 is a good choice :) )

rlberry/envs/tests/test_bandits.py
```diff
@@ -1,20 +1,35 @@
...
     assert np.abs(np.mean(sample) - 1) < 0.1
 
 
 def test_cor_normal():
     env = CorruptedNormalBandit(means=[0, 1], cor_prop=0.1)
-    safe_reseed(env, Seeder(42))
+    safe_reseed(env, Seeder(TEST_SEED))
 
     sample = [env.step(1)[1] for f in range(1000)]
     assert np.abs(np.median(sample) - 1) < 0.5
```
TimotheeMathieu, line 35: Maybe we can also include some generic bandit testing using something like in PR #110 ? You can do it here or I will do a separate PR.

rlberry/agents/bandits/index_agents.py
```diff
@@ -34,9 +34,8 @@ class IndexAgent(BanditWithSimplePolicy):
 
     name = "IndexAgent"
 
-    def __init__(self, env, index_function=lambda rew, t: np.mean(rew), **kwargs):
+    def __init__(self, env, index_function=None, **kwargs):
         BanditWithSimplePolicy.__init__(self, env, **kwargs)
-        self.n_arms = self.env.action_space.n
```
TimotheeMathieu, line 39: If you rename all the "index" to "indices", shouldn't it be "indices_function" and "Indice_Agent" ?

rlberry/agents/bandits/randomized_agents.py
```diff
@@ -0,0 +1,122 @@
...
+        See :class:`~rlberry.envs.bandits.Bandit`.
+
+    index_function : callable or None, default = None
+        Compute the index for an arm using the past rewards and sampling
+        probability on this arm and the current time t.
+        If None, use loss-based importance weighted estimator.
+
+    prob_function : callable or None, default = None
+        Compute the sampling probability for an arm using its index.
+        If None, use exponential softmax probabilities with any learning rate.
```
TimotheeMathieu, line 24: I don't understand the bit about "with any learning rate", can you explain more. Maybe you can say "with learning rate from ..." and cite an article.

TimotheeMathieu (commented):
Thank you for this and thank you for the changes to the bandit tests. 
I have some comments and otherwise I also have a question. If I understand right, Exp3 is about adversarial bandits. Would it be possible to have some sort of "adversarial environment" and do an example on it ? That would be nice and that would explain why Exp3. I think one important thing for bandits in rlberry is also to provide some benchmark environments for difficult cases (adversarial, outliers, privacy...) on which the community is currently doing some research.

rlberry/utils/check_bandit_agent.py
sauxpa: Agreed, done :)

rlberry/envs/tests/test_bandits.py
sauxpa: Agreed, I'll have a look. Maybe a separate PR for clarity? Ideally it would be great to agree on a good set of tests for bandits and just copy-paste those when we implement more agents, so I think it is worth dedicating some time on this

rlberry/agents/bandits/randomized_agents.py
sauxpa: It's just bad writing from me. I'll resubmit (need to make a small change to the learning rate definition actually -- there are quite a few in the literature, I'm trying to pick a somewhat standard one).

rlberry/manager/evaluation.py
sauxpa: Absolutely, when it's out I'd much rather see quantiles (the update is already on github btw, just not released yet to the latest version of seaborn).

rlberry/envs/tests/test_bandits.py
TimotheeMathieu: Ok. Ideally, the generic check_bandit function should test only basics properties (reproducibility, save & load and things like that) but maybe we want to keeep some specific tests for each bandit, for instance test that it works well on subgaussian and not only on bernoulli, test that adversarial bandits work on adversarial problems...

rlberry/agents/bandits/index_agents.py
TimotheeMathieu: Ok, maybe I should read the Oxford dictionary some more ;)

sauxpa:
@TimotheeMathieu : I have given some thoughts on implementing an "adversarial" environment, below is a short recap for both of us.

* The "true" adversarial environment would be one where the user is asked to manually provide the rewards, either from the start (oblivious opponent) or actively during the run of the algorithm (adaptive opponent). 
* Many paper in this literature run experiments on non-stationary rather truly adversarial environments, where typically one arm is the best for many rounds and then becomes the worst (see e.g the experiments in the Tsallis-INF paper).
* Another related problem is that of minimax lower bounds: the supremum of regret at horizon T over a certain family of bandit instances is attained for a certain value of the gap, typically O(\sqrt{1/T}). We might want to have some experiments on it also, but I'll skip it for now.

What I plan is to add an oblivious opponent environment. The adaptive one is more cumbersome and not really useful IMO.
That would give us already much flexibility to design interesting examples. Adding non-stationary bandits in rlberry is a whole topic in itself, and eventually I think we will want to have dedicated environments for them too, but in the meantime we should be able to reproduce standard experiments in the adversarial literature.

TimotheeMathieu (approved):
LGTM, thanks.

Issue #121, opened 2022-02-16
Minimize star imports


mmcenta:
We should minimize star imports (that is, using 'from ... import *) because:

1) They reduce code readability (it becomes harder to pinpoint from where an object was imported);

2) They can cause bugs when a star import overrides an object previously imported. These are typically difficult to catch.

3) They are actually discouraged in most styleguides.

_As suggested in https://github.com/rlberry-py/rlberry/pull/110#discussion_r807213843_

TimotheeMathieu:
We can also test for them in the "formating check" that we do at each PR to ensure that we respect this guideline, flake8  (something like F405 that I can include in the check).

mmcenta:
I agree with the formatting check.

I'm assigning myself to this issue and I will submit a PR to correct the style soon!

PR #125, opened 2022-02-18
Minimize star imports

mmcenta:
A quick fix to #121.

I left the star imports on `rlberry/envs/bandits/__init__.py` and `rlberry/tests/test_agent.py` because they are importing multiple things while also being quite clear. Let me know what you think.

@TimotheeMathieu do you know how to set up the checks in flake8?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -57,7 +57,7 @@ jobs:
     displayName: "black"
   - script: |
       # ensure there is no unused imports with
-      flake8 --select F401 --exclude=rlberry/check_packages.py --per-file-ignores="__init__.py:F401"
+      flake8 --select F401,F405 --exclude=rlberry/check_packages.py --per-file-ignores="__init__.py:F401"
     displayName: 'flake8'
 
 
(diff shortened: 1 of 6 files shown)


TimotheeMathieu:
and you can check that you did not forget any star import using the command 
`flake8 --select F405,F401 --exclude=rlberry/check_packages.py --per-file-ignores="__init__.py:F401"`

TimotheeMathieu:
I think it is better to also check this file now that it has become a general guideline in the repo.

mmcenta:
You're right, I just wanted to avoid importing all agents by hand but I think I figured out a solution:
```python
# Old star import (fails F405)
from rlberry.agents import *
from rlberry.agents.torch import *

agents = [ValueIterationAgent, PPOAgent, ...]

# New syntax
import rlberry.agents as agents
import rlberry.agents.torch as torch_agents

agents = [agents.ValueIterationAgent, torch_agents.PPOAgent, ...]
```
It is a bit more verbose but it solves our problems.

PR #126, opened 2022-02-19
feat (reproducibility)

TimotheeMathieu:
…nager

### DIFF
--- a/rlberry/manager/agent_manager.py
+++ b/rlberry/manager/agent_manager.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from rlberry.seeding import safe_reseed, set_external_seed
 from rlberry.seeding import Seeder
 from rlberry import metadata_utils
+from rlberry.utils.hash_utils import get_git_hash, get_pickle_md5
 
 import functools
 import json
@@ -118,6 +119,7 @@ class AgentHandler:
         """Saves agent to file and remove it from memory."""
         if self._agent_instance is not None:
             saved_filename = self._agent_instance.save(self._fname)
+
             # saved_filename might have appended the correct extension, for instance,
             # so self._fname must be updated.
             if not saved_filename:
@@ -125,7 +127,9 @@ class AgentHandler:
                     f"Instance of {self._agent_class} cannot be saved and will be kept in memory."
                 )
                 return
+
             self._fname = Path(saved_filename)
+
             del self._agent_instance
             self._agent_instance = None
 
@@ -700,6 +704,25 @@ class AgentManager:
             except Exception as ex:
                 logger.warning("[AgentManager] Instance cannot be pickled: " + str(ex))
 
+        # save agent info
+        saved_filename_id = str(filename) + "_id.json"
+        agent_hash = get_pickle_md5(self.agent_class)
+        init_kwargs_copy = deepcopy(self.init_kwargs)[0]
+        del init_kwargs_copy[
+            "seeder"
+        ]  # we don't save the seeder as it is expected to change
+        del init_kwargs_copy[
+            "output_dir"
+        ]  # we don't save the output dir as it changes with the date.
+        kwargs_hash = get_pickle_md5(init_kwargs_copy)
+        rlberry_ver = get_git_hash()
+        data = {
+            "agent_hash": agent_hash,
+            "kwargs_hash": kwargs_hash,
+            "rlberry_commit": rlberry_ver,
+        }
+        _safe_serialize_json(data, Path(saved_filename_id))
+
         return filename
 
     @classmethod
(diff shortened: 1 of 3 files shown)


omardrwch:
I'm not sure I understand why you need the `load_instanciated` method, what is your use case?

To check equality between two `AgentManager` instances, we could override the `__eq__` method in `AgentManager`, instead of defining this new `load_instanciated` method (which I find a bit weird).
An example of what we could have:

```python
params = dict(...)
params_alternative = dict(...)

manager1 = AgentManager(AgentClass, env,  init_kwargs=params)
manager2 = AgentManager(AgentClass, env,  init_kwargs=params)
manager3 = AgentManager(AgentClassAlternative, env,  init_kwargs=params_alternative)

# override __eq__ such that:
manager1 == manager2      #  True
manager1 == manager3      #  False
```

TimotheeMathieu:
The use case is : I ran an experiment last year, using rlberry v0.2.1 and a custom agent and custom hyperparameters. Now, I want to reuse the files that were automatically saved without running the experiment, typically to compare to a new agent. How do I check that the agent is indeed the agent I want and that the rlberry version is indeed the same version ? How do I check that I have indeed the right experiment ?
If I use the AgentManager.load, I can't check that the hyperparameter and agent I propose are indeed the ones I am thinking about. Instead, the if I instantiate the agent with all the right parameters and with the right code, I can do the check. Typically, I would use the exact same script as the one used to run the experiment last year but I check that the agent did not change during the year.

I want to compare an agent with the agent I load so I would nonetheless have to load the agent but yes, I could maybe override the equality... but the idea is to have something that is transparent. You don't have to do anything, rlberry will automatically say whether you loaded a file that correspond to the experiment you define. 

Is that clear ? It is related to the reproducibility of results and dump of agent we were talking about in the last WG.

omardrwch:
If I understood correctly, you have the same script for running experiments and for plotting, and a flag `fit = True / False` to indicate whether you want to train or just plot.

In my experiments, I usually have one script for running the agents, and another script for plot only. In this case, there is no issue of making sure that your plot corresponds to the agent that you ran before. All information for reproducibility is in the folder where the experiments where saved (we can store the rlberry version, hyperparameter, the hash identifying the last git commit etc.).

If you prefer to use a single script, I still think it is better to override `__eq__` since, for me, it feels very unnatural to create an instance of `AgentManager` with a set of parameters and allow it to load another manager with other parameters. The purpose of `load_instanciated`   seems to be generating a warning only, so we can allow the user to check the equality of managers/experiments directly with `__eq__` instead.

The `__eq__` approach can also be useful when we have two folders with two sets of results, and we wonder whether the experiment was the same.

So what do you think of the following for your use case (let me know if I didn't understand correctly):

```python
plot_only = True
fname = 'my_experiment.pickle'

manager = AgentManager(...)

if plot_only:
    loaded_manager = AgentManager.load(fname)
    assert manager == loaded_manager, "experiments must be the same!"
    manager = loaded_manager
else:
    manager.fit()
    saved_fname = manager.save()
    print(saved_fname)

# plot etc
plot_writer_data(manager...)
```
?

TimotheeMathieu:
Yes your proposition is nice, I will do this (i.e. overwrite the `__eq__`).

TimotheeMathieu:
I did it, it was fairly simple. 
Should I still keep the hash and the git commit utils ? I don't know if they can be useful. Maybe in deep learning when we don't want to dump the whole network but still we want to retrieve its writer and compare the agents.

rlberry/manager/agent_manager.py
```diff
@@ -729,8 +732,20 @@ def load(cls, filename):
 
         obj.__dict__.clear()
         obj.__dict__.update(tmp_dict)
+
         return obj
 
+    def __eq__(self, other):
+        self_agent_init_kwargs = deepcopy(self.init_kwargs)[0]
```
omardrwch, line 739: I think we should also check the equality of `eval_kwargs` and `fit_kwargs`

omardrwch (changes requested):

omardrwch:
> I did it, it was fairly simple. Should I still keep the hash and the git commit utils ? I don't know if they can be useful. Maybe in deep learning when we don't want to dump the whole network but still we want to retrieve its writer and compare the agents.

I like the git commit utils, maybe we could store the hash of the current commit as an attribute of `AgentManager`, what do you think?  For the hashing of the object, if we're not using `get_object_md5` for now, I think it's ok to remove it now and add it later if needed (since it's only one line)

TimotheeMathieu:
I wonder about this because I think the git commit thing would not work with a released version (i.e. pip install rlberry). Maybe we should have some `get_rlberry_version` that returns the git commit if available or return rlberry version if git commit is not available... 

I will look into it. This is a bit of a pain because I will have to install rlberry in non-editable mode to check out how pip install rlberry outside of git repo to check my function. I will do that with the new rlberry container I constructed this morning :)

omardrwch:
Good idea, having a `get_rlberry_version` function that does that would be nice. If you prefer, we can do that later in another PR.

TimotheeMathieu:
No I think this PR is ok for this. I will just change the name.

rlberry/manager/agent_manager.py
TimotheeMathieu: The advantage of my hash approach is that I tested everything (i.e. even new attributes that the user may have defined in AgentManager, it also tested eval env, train env... everything). But maybe it is not really necessary.

omardrwch: If needed, I think the hash approach can also be done in `__eq__`, no? Something like `hash(self) == hash(other)`

TimotheeMathieu: I think let us go for the simple solution. If people really want to be reproducible they would have to use container or something like that and not only this kind of software level check. I will keep it simple and only test the `{eval,fit,init}_kwargs` and the agent_class this should be sufficient.

TimotheeMathieu: Done.

TimotheeMathieu:
@omardrwch : now, AgentManager record rlberry version. If git is not available, use rlberry version. To do that, I had to define `rlberry.__version__` and I changed the version everywhere for "0.3.0dev0" because I think 0.3.0 should be the next version ?

omardrwch (approved):

PR #127, opened 2022-02-20
(fix) bug azure pipeline windows

TimotheeMathieu:
Fix failing check on windows azure pipeline due to some error with numpy. Linked to PR #124

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,8 +10,8 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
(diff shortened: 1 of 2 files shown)


TimotheeMathieu:
The problem seems to be with the new update of `pyopengl` that happened yesterday. For now, I solve it by specifying the good version in `requirements.txt`, i.e. `PyOpenGL==3.1.5` and `PyOpenGL_accelerate==3.1.5` but later on when (hopefully) a patch is submitted, we should reset requirements to any version.

PR #132, opened 2022-02-24
[MRG] (feat) Readme update

TimotheeMathieu:
Update the readme to remove dead links and redundant info with doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Install the latest version for a stable release.
 $ pip install -U git+https://github.com/rlberry-py/rlberry.git@v0.2.1#egg=rlberry[default]
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -70,13 +71,21 @@ environments etc etc!
...
-Check our [getting started section](#getting-started)!
+## Installation
+
+Install the latest version for a stable release.
+
+```bash
+$ pip install -U git+https://github.com/rlberry-py/rlberry.git@v0.2.1#egg=rlberry[default]
+```
+
+rlberry dependencies are listed in [requirements.txt](https://github.com/rlberry-py/rlberry/blob/main/requirements.txt). The documentation includes more [installation instructions](https://rlberry.readthedocs.io/en/latest/installation.html) in particular for users that work with Jax.
```
omardrwch, line 82: The dependencies depend on the installation mode, right? The [default] mode does not include pytorch, for example, but torch is in `requirements.txt`

omardrwch (approved):

Issue #134, opened 2022-03-01
Bug when creating AgentManagers


mmcenta:
Hello, I was recently working with the agent checks on `rlberry/utils/check_agent.py` and I was encountering the following error:

>  fatal: not a git repository (or any of the parent directories): .git
Traceback (most recent call last):
  File "./examples/demo_examples/demo_sb_agent.py", line 30, in <module>
    stats = AgentManager(
  File "/Users/mmcenta/miniconda3/envs/rlb/lib/python3.8/site-packages/rlberry/manager/agent_manager.py", line 386, in __init__
    self.rlberry_version = get_rlberry_version()
  File "/Users/mmcenta/miniconda3/envs/rlb/lib/python3.8/site-packages/rlberry/utils/hash_utils.py", line 25, in get_rlberry_version
    subprocess.check_output(
  File "/Users/mmcenta/miniconda3/envs/rlb/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/Users/mmcenta/miniconda3/envs/rlb/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'rev-parse', 'HEAD']' returned non-zero exit status 128.

It is caused by the `get_rlberry_version()` function introduced recently (in PR #126), which requires the current directory (or its parent) to be a  git repository. As it turns out, after installing the library with `pip install .`, this function is called from inside the library (somewhere in `~/miniconda3/envs/rlb/lib/python3.8/site-packages/rlberry/utils`, in my case). The result is the error above because the other directory is not part of a git repository.

### Reproducing

The code below reproduces the bug:

```python
from rlberry.agents.torch import A2CAgent
from rlberry.envs import gym_make
from rlberry.manager import AgentManager

if __name__ == "__main__":
    env_ctor = gym_make
    env_kwargs = {'id': 'CartPole-v0'}
    stats = AgentManager(
        A2CAgent,
        (env_ctor, env_kwargs),
        agent_name="A2C",
        init_kwargs=dict(batch_size=16),
        fit_budget=100,
    )
```

Running this code produces the same error from above. Do NOT run the code from the `rlberry` main directory, because in that case it will import the local code which runs in a git directory. I run this code from a subdirectory `rlberry/test_bug/bug.py`, which forces Python to load the installed version of rlberry.

[...]

Issue #133, opened 2022-02-24
Wrapper agents are not compatible with AgentManager/check_rl_agent


mmcenta:
I've been working on an agent that wraps StableBaselines3 algorithms, but I've been having issues trying to use it with `AgentManager`/`check_rl_agent`. What I currently have looks something like this:

```python
class StableBaselinesAgent(AgentWithSimplePolicy):
    """ Wraps an StableBaselines3 Algorithm with a rlberry Agent.

    Notes
    -----
    Other keyword arguments are passed to the algorithm's constructor.

    Parameters
    -----------
    env: gym.Env
        Environment
    algo_cls: stable_baselines3 Algorithm class
        Class of the algorithm to wrap (e.g. A2C)
    policy: str or stable_baselines3 Policy class
        Policy to use (e.g. MlpPolicy)
    """
    def __init__(
        self,
        env: types.Env,
        algo_cls: Type[SB3Algorithm],
        policy: Union[str, Type[SB3Policy]],
        **kwargs,
    ):
        super(StableBaselinesAgent, self).__init__(env, **kwargs)
        ...

# Example Usage
from stable_baselines3.a2c import A2C
agent = StableBaselinesAgent(A2C, 'MlpPolicy', env_ctor(**env_kwargs))
agent.fit(budget=1000)
```

However, both `AgentManager` and `check_rl_agent` take *agent classes* as input. 

## Another example

Suppose that you want to extend the functionalities of some agents. The usual way to approach this problem is to subclass the original agents (e.g. DQN) as follows:

```python
class BetterDQN(DQN):
    def __init__(self, env, **kwargs):
        ...

    def additional_func(self, *args):
        ...
```

If you want to create a general wrapper that works for any agent, you would have something like this:

```python
class BetterAgent(Agent):
    def __init__(self, env, agent_cls, **kwargs):
        ...

    def additional_func(self, *args):
        ...
```

which is very similar to the situation I encountered. This has the potential to be even worse to deal with if we stack multiple wrappers.

## Solution 1: `init_kwargs`
This could be solved with the`init_kwargs` argument:

```python
stats = AgentManager(
    StableBaselinesAgent,
    (env_ctor, env_kwargs),
    agent_name='A2C',
    init_kwargs=dict(algo_cls=A2C, policy='MlpPolicy', verbose=1),
    ...
)
```

[...]

PR #135, opened 2022-03-01
[MRG] Support `init_kwargs` in agent checks

mmcenta:
Good afternoon!

As discussed in #133, we opted to add support for the `init_kwargs` argument for AgentManagers in the agent checks available at `rlberry/utils/check_agent.py`. The implementation is pretty straightforward, but I would like to mention two things:

1. I changed some variable names to conform to snake_case.
2. I discovered a bug when creating AgentManagers due to a recent reproducibility feature (issue #134).

Let me know what you think!

### DIFF
--- a/rlberry/utils/check_agent.py
+++ b/rlberry/utils/check_agent.py
@@ -49,8 +49,9 @@ def _fit_agent_manager(agent, env="continuous_state", init_kwargs=None):
 
     train_env = _make_env(env)
     try:
-        agent = AgentManager(agent, train_env, fit_budget=5, n_fit=1, seed=SEED,
-            init_kwargs=init_kwargs)
+        agent = AgentManager(
+            agent, train_env, fit_budget=5, n_fit=1, seed=SEED, init_kwargs=init_kwargs
+        )
         agent.fit()
     except Exception as exc:
         raise RuntimeError("Agent not compatible with Agent Manager") from exc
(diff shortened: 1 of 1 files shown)


mmcenta:
@TimotheeMathieu @omardrwch Do you have any idea why the checks are failing here?

TimotheeMathieu:
Do you have the last version of black ? it should be version 22.1.0

mmcenta:
My bash script was only running flake8 and not black, that one is my fault.

TimotheeMathieu:
Just a remark : you could use pre-commit. I've set up a pre-commit config in rlberry and it works pretty well for me. Although it checks a bit more than what is necessary for a green PR check.

mmcenta:
I will look into setting that up!

mmcenta:
> Just a remark : you could use pre-commit. I've set up a pre-commit config in rlberry and it works pretty well for me. Although it checks a bit more than what is necessary for a green PR check.



> Just a remark : you could use pre-commit. I've set up a pre-commit config in rlberry and it works pretty well for me. Although it checks a bit more than what is necessary for a green PR check.

Thanks, I will check it out!

TimotheeMathieu (approved):
LGTM. just a question though, do we want to be snake_case compliant ? I did not pay attention to that before but maybe we should.

mmcenta:
> LGTM. just a question though, do we want to be snake_case compliant ? I did not pay attention to that before but maybe we should.

PEP8 recommends snake_case for variables (https://www.python.org/dev/peps/pep-0008/#naming-conventions).

mmcenta:
@TimotheeMathieu Maybe I'm out of the loop, but is there a preferable way to merge PRs into rlberry? The last time I created a merge commit but some people prefer squash/rebase merges. Have we discussed this issue already?

TimotheeMathieu:
I personally prefer squash as I find it easier to navigate in history if there is one commit for one feature/PR but I don't think we have any general rule so if you prefer merge commit go ahead.

PR #136, opened 2022-03-01
(fix) fix get_version_rlberry git commit

TimotheeMathieu:


TimotheeMathieu:
Ok. I think this should work, but I have to merge to see if this work.

The idea is that now the version is in a static file in ` rlberry/_version.py` and there is a github action that update this version every time a pull request is merged. 
This should work.
Let us test it.

PR #137, opened 2022-03-01
[WIP] Template file to define indices in bandits

sauxpa:
UCB, MOSS, KL-UCB, IMED...

### DIFF
--- a/examples/demo_bandits/plot_compare_index_bandits.py
+++ b/examples/demo_bandits/plot_compare_index_bandits.py
@@ -7,11 +7,17 @@ This script Compare several bandits agents and as a sub-product also shows
 how to use subplots in with `plot_writer_data`
 """
 import numpy as np
-from rlberry.envs.bandits import BernoulliBandit
-from rlberry.agents.bandits import IndexAgent, RandomizedAgent
-from rlberry.manager import AgentManager, plot_writer_data
 import matplotlib.pyplot as plt
+from rlberry.envs.bandits import BernoulliBandit
+from rlberry.manager import AgentManager, plot_writer_data
 from rlberry.wrappers import WriterWrapper
+from rlberry.agents.bandits import (
+    IndexAgent,
+    RandomizedAgent,
+    makeBoundedUCBIndex,
+    makeETCIndex,
+    makeBoundedMOSSIndex,
+)
 
 # Agents definition
 # sphinx_gallery_thumbnail_number = 2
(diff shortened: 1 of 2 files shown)


TimotheeMathieu:
After reflexion

-  I am ok with the Bandit tracker, but it must be documented to explain exactly how things are done in rlberry for bandits. I think a tuto bandit and maybe some more docstring on `BanditTracker`, in particular we need a more thorough doc on the interaction between the different classes. Also the docstring of `BanditTracker` has a typo: it is not n_arms anymore, it is agent in the parameters.  Ideally the tracker should not be mandatory to code a bandit in rlberry but it should make things a lot easier.
- Maybe an API similar to a writer API would be easier ? i.e. implementing a method `add_scalar_tracker` or something similar to modify an existing tracker, adding a new thing to track ? I think this would make it easier to customize what is tracked.
- We discuss in the Working Group and one (or two) level of abstraction is OK, so we do not need everything in one file. We try to keep it simple but if some abstraction is helpful, let us do it: maybe have a folder `tools` in the bandit folder and as tools, we would have the `make_index` and the tracker. And later tools tailored for instance to linear bandits, to privacy in bandits, to robustness in bandits...

sauxpa:
- BanditTracker has been moved to a separate **tools** folder
- It is now inherited from DefaultWriter instead of rewriting some read/write logic specifically for bandits (en passant, I have added a few helper functions to DefaultWriter)
- The way to access the data stored in the tracker is through the read_last_tag_value or self.read_tag_value methods of DefaultWriter. It makes defining the indices a bit verbose, I might add some extra API to have a more concise call for some attributes like n_pulls; I leave it like this for now to avoid adding more layers of code.
- I agree more docs and tuto are required. This is my next TODO.
- Also: would need to clean up TS. I'll see if I just move the posterior update logic to the tracker or if it needs further rewriting (support more prior, more modular design). For the EoW target, maybe just the former.

TimotheeMathieu (approved):
Looks god to me, thanks ! 
Indeed, this may need a bit more documentation. Anyway I plan to do a quickstart for bandit users sometime soon. We could also add a example section in the docstring (as it is done for thompson sampling)  in particular for BanditTracker to show a typical example of BanditTracker. This may be done in a separate PR.

sauxpa:
> Looks good to me, thanks ! Indeed, this may need a bit more documentation. Anyway I plan to do a quickstart for bandit users sometime soon. We could also add a example section in the docstring (as it is done for thompson sampling) in particular for BanditTracker to show a typical example of BanditTracker. This may be done in a separate PR.

Cool, I'll merge tomorrow

PR #144, opened 2022-03-02
feat (agent and env) add get_params and set_params

TimotheeMathieu:
Copy of scikit-learn API get_params and set_params.

Example of use:

```
>>> env = Chain(fail_prob=0.8)
>>> env.get_params()
{'L': 5, 'fail_prob': 0.8}
>>> env.set_params(fail_prob=0.5)
>>> env.get_params()
{'L': 5, 'fail_prob': 0.5}
>>> env_ctor = env.__class__
>>> env_kwargs = env.get_params()
```

I implemented this in Agent and Model base classes. For these functions to work, all the parameters given to __init__ should become self.param. I changed all the agents and environments in rlberry to respect this rule. I also implemented a check/test for this and in the idea it should be an optional feature in the sense that if a user's custom agent does not support get_params, the main rlberry functionality should still work but there may be some missing small features. On the other hand all rlberry agents/env should support it ? (can be subject to discussion)

Related to a problem discussed with @omardrwch  and @riccardodv , with this we can properly construct the (ctor, kwargs) from a built env or agent.

### DIFF
--- a/rlberry/agents/agent.py
+++ b/rlberry/agents/agent.py
@@ -12,6 +12,8 @@ from rlberry.seeding import safe_reseed
 from rlberry.envs.utils import process_env
 from rlberry.utils.writers import DefaultWriter
 from typing import Optional
+import inspect
+from collections import defaultdict
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


xuedong:
Hi, just by curiosity, why is set_params removed?

omardrwch:
> Hi, just by curiosity, why is set_params removed?

Because setting and agent's parameters might require it to be initialized again (with __init__) and maybe create unexpected behaviors (if the different objects used by the agent were constructed based on the parameter's values). So I suggested to remove set_params, and re-create the objects when we need to change its parameters.

rlberry/agents/agent.py
```diff
@@ -302,6 +303,51 @@ def load(cls, filename, **kwargs):
...
+            if p.name != "self" and p.kind != p.VAR_KEYWORD
+        ]
+
+        # Extract and sort argument names excluding 'self'
+        return sorted([p.name for p in parameters])
+
+    def get_params(self, deep=True):
+        """
+        Get parameters for this agent.
+        Parameters
```
xuedong, line 332: A very minor comment on the style of docstrings, I see in other files we have blank lines to separate different sections (Parameters, Returns, References, etc), e.g. in other files a blank line would be put on line 332. Shall we keep the same standard for all files?

rlberry/utils/check_env.py
```diff
@@ -35,3 +35,21 @@ def check_env(env):
...
+
+def check_rlberry_env(env):
+    """
+    Companion to check_env, contains additional tests. It is not mandatory
+    for an environment to satisfy this check but satisfying this check give access to
+    additional features in rlberry.
+
+    Parameters
+    ----------
+
```
xuedong, line 48: Same minor comment here: shall we remove the blank line on line 48?

xuedong (approved):
Hi, apart from some very minor comments, I think the PR is ready.

rlberry/agents/agent.py
TimotheeMathieu: I don't really have an opinion on this. I searched in numpy docstring and pep rules but I only found this in https://www.pydocstyle.org/en/2.1.1/error_codes.html . 
We may want to check this in the check if we really want to enforce this, but I think that as long as the new header is detected by sphinx in the doc, I  am happy with or without the blank lines.

TimotheeMathieu: We could use this : https://pypi.org/project/flake8-docstrings/ to integrate this into flake8 format check.

TimotheeMathieu:
For now, the new check I have only check numpy docstring style. In order to check these blank lines, we would need to check pep257. 
Do we want to ask for pep257 ? If yes, I would also change this into the contributing guideline.

xuedong:
> 

I would recommend yes.

TimotheeMathieu:
Ok done, for now I check only the mentioned D410 to D412 errors (i.e. blank lines) because there are a lot of errors if we check all of flake8-docstring. i will make an issue of it that we may or may not have the courage of solving in a foreseeable futur.

xuedong (approved):

omardrwch (approved):

PR #148, opened 2022-03-03
[MRG] StableBaselines Agent

mmcenta:
I'm finishing work on the StableBaselines wrapper that I was working on before. Unfortunately, SB3 algorithms don't seem to reliably satisfy `check_fit_additive`, so I removed it from the tests.

Let me know what you think!


mmcenta:
It seems tests are failing because `stable_baselines` isn't a requirement. Do we set up an optional requirement such as with torch?

TimotheeMathieu:
Yes I think it is ok to add it as optional requirement because I think the people that install torch will also want to install stablebaseline. 
My only concern is about the install time when doing the tests but I guess it should not be a huge increase because a lot of stablebaselines's dependencies are already dependencies of rlberry.

mmcenta:
@omardrwch mentioned that maybe we can create a dummy SB3 algorithm class and avoid the import issues. I will look into this.

rlberry/utils/tests/test_sb_agent.py
```diff
@@ -0,0 +1,37 @@
...
+def _sb3_check_rl_agent(agent_cls, env, init_kwargs=None):
+    """
+    Check that an RL agent can be instantiated and fit.
+
+    Note
+    ----
+    Does not check that the fit is additive, because that doesn't seem to be
+    always the case with StableBaselines3.
+    """
+    check_agent_manager(agent_cls, env, init_kwargs=init_kwargs)
```
omardrwch, line 21: Maybe we could increase the `verbose` in the `init_kwargs` here to make sure that the writer wrapper is tested?

omardrwch (changes requested):

rlberry/utils/tests/test_sb_agent.py
mmcenta: I'm taking a look into this. There's something wrong with the loading when I set `verbose=1`.

rlberry/utils/tests/test_sb_agent.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def _sb3_check_rl_agent(agent_cls, env, init_kwargs=None):
+    """
+    Check that an RL agent can be instantiated and fit.
+
+    Note
+    ----
+    Does not check that the fit is additive, because that doesn't seem to be
+    always the case with StableBaselines3.
```
TimotheeMathieu, line 21: this seems to be due to the callback that can stop the fit early (e.g. https://stable-baselines.readthedocs.io/en/master/guide/callbacks.html `_on_step` method). Would it make sense to redefine the default stablebaselines callback in order to avoid stopping early ? And is this early stopping a problem for Hyperparameter optimization ? (also @omardrwch on this)

TimotheeMathieu (commented):
Thanks, this looks nice !
I have just a small comment on the test and some comment on the doc.
Concerning the doc, could you move the link to the stablebaselines tuto (and also the link to gym tuto) from the index page to the user guide ? I think it would make more sense there. Also could you add the wrapper in the API (file docs/api.rst) ?

mmcenta:
On another note, I added a test to test `tensorboard` integration but the test seems to be failing when I try to import tensorboard. Should I remove it?

docs/api.rst
```diff
@@ -134,6 +134,14 @@ Seeding
 Utilities, Logging & Typing
 ====================
 
+StableBaselines3 Agent
```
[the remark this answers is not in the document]
mmcenta: I agree that it is better to put it under Agents, but it is currently under `rlberry/utils`, should I move it to the `agents` directory too?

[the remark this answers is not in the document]
TimotheeMathieu: Yes I think it make sense to have this in `agents` directory. Even though it is a tool, the use case is people use it almost as an agent (in the tests for instance you use it as an agent that takes as argument the sb class).

rlberry/utils/tests/test_sb_agent.py
TimotheeMathieu: After looking into it, I think the default is no early stopping in stablebaselines (I had to go back several inherited modules in stablebaselines, this is quite the maze). The problem seems to be seomewhere else.

On the other hand, the learn method has  a loop that says 
`for update in range(1, total_timesteps // self.n_batch + 1):` which would not give a total of  `total_timesteps` iterations unless `self.n_batch` divides `total_timesteps`. Maybe the problem comes from here ?

mmcenta:
@TimotheeMathieu On another note, I think PR #144 broke my git pre-commit setup - now flake8 stops me because of the unused imports on the `__init__.py` files that used to be ignored. Do you know how to fix this?

docs/api.rst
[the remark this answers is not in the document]
mmcenta: I just pushed the changes that move it into `agents`, let me know what you think.

docs/api.rst
```diff
@@ -119,6 +118,14 @@ Importation tools
...
     envs.gym_make
 
+StableBaselines3 Agent
+-----------------------
+.. autosummary::
+   :toctree: generated/
+   :template: class.rst
+
+   agents.StableBaselinesAgent
+
```
TimotheeMathieu, line 128: I would have moved this to line 51 with something like 

```
   agents.torch.AVECPPOAgent
   agents.torch.REINFORCEAgent

 Importation tools
-----------------------
.. autosummary::
   :toctree: generated/
   :template: class.rst

   agents.StableBaselinesAgent
``` 
Jax Agents (experimental)
--------------------

mmcenta: Maybe we should put it before the Torch agents for now (since they'll be released as experimental). What do you think?

mmcenta:
> I don't really know what we should do. i mean, tensorboard is not huge (5Mb) so I think we could just add it to the requirements and be done with it. On the other hand, if we want to stay compatible with everyone we may encounter this problem again in the futur and it would not be good to always add requirements like that. One solution would be to have a directory for additional "extra" tests, these tests we could check on a parallel check of azure pipeline (have a pipeline on linux that checks only these tests in order not to have a too long running time for the checks, with its own additional dependencies that would not appear in requirements.txt). Or we could also not test them at all, but still I think it is good to have these tests so we keep them, the only question is do we check them in azure pipeline or not ?

@omardrwch What do you think about this? I'm tempted to just add it to the requirements for now.

Also, I'll look into the `batch_size` problem that @TimotheeMathieu talked about. Maybe we can set `batch_size=1` also solve the issues I was having, but I'll have to run some tests again.

docs/api.rst
TimotheeMathieu: Yes you are right.

mmcenta: Also, about the pre-commit you're right - it was caused by me updating one of the `__init__.py`'s, but I used `--no-verify` to commit anyways. We can create another PR to add the ignores later.

TimotheeMathieu:
I did some experiments and I think it may be because of `reset_num_timesteps`. With the following fit function:
```
    def fit(
        self,
        budget: int,
        tb_log_name: Optional[str] = None,
        reset_num_timesteps: bool = True,
        **kwargs,
    ):
        """Fit the agent.

        Note
        ----
        This method wraps the :code:`learn` method of the algorithm.
        Logging parameters are processered by rlberry in order to use the
        agent.writer.

        Parameters
        ----------
        budget: int
            Number of timesteps to train the agent for.
        tb_log_name: str
            Name of the log to use in tensorboard.
        reset_num_timesteps: bool
            Whether to reset or not the :code: `num_timesteps` attribute
        """
        # if not self.wrapped._custom_logger:
        #     if tb_log_name is None:
        #         tb_log_name = self.wrapped.__class__.__name__
        #     sb_logger = configure_logger(
        #         self._verbose, self._tb_log, tb_log_name, reset_num_timesteps
        #     )
        #     self.set_logger(sb_logger)
        self.wrapped.learn(total_timesteps=budget, reset_num_timesteps=False, **kwargs)
```
I obtain the following add graph:

results_add_0

and I guess this is OK, with batch size 1 the fit is additive (which is what we want).

mmcenta:
@TimotheeMathieu 
I think you're right: I assume that `reset_num_timesteps` is used only for logging purposes (as they mention in their docs), but it seems that it is used to possibly reset the environment too.
However, I can't comment out that part of the code because that's what integrates their logging to rlberry. I came up with a solution that seems to solve the issue - I'll push the changes as soon as I'm finished with testing.

mmcenta:
Just an update: I've been able to (partially) fix the additive fit issues for the agent. 

I encountered one last problem: it seems that it is difficult to guarantee additive fits when a model is saved and then loaded during training. First, the standard `.save()` and `.load()` methods from StableBaselines don't support additive fits by themselves. By using `.get_parameters()` and `.set_parameters()` instead, I can get the additive fits to work in isolation, but the method breaks when I implement it in the agent - although it does seem some sort of approximation error since the predicted values are very close.

This is relevant because our `AgentManager` saves and loads agents all the time, so I can't guarantee additive fits in that case. More specifically, our additive fit test uses `AgentManager`'s, so it still fails :(

omardrwch:
> @omardrwch What do you think about this? I'm tempted to just add it to the requirements for now.

Sorry, I'm a bit (very) late in this PR. Regarding the tensorboard requirement, I think it's ok to add it for now. I think the SB3 wrapper is a very important feature, so testing it should be worth the extra requirement. Later we could implement @TimotheeMathieu 's suggestion (of creating a directory for additional heavier tests) when needed (e.g. also for a more comprehensive testing of our agents, to make sure that they match some benchmarks).

TimotheeMathieu (approved):
Ok for me. We can make separate PRs for the tests and seeder later. Thanks.

PR #149, opened 2022-03-07
[MRG] feat (bandits) Real bandits experiment and basic BAI agent

TimotheeMathieu:
This is an example. Not part of the main API.

I implement an environment for a  real bandit experiment in which I search for the fastest mirror among a list of mirrors to download ubuntu. This can serve as a generator of real data to test a bandit algorithm.

I implement a sequential halving agent for a basic best arm identification algorithm.


TimotheeMathieu:
Does not change the API. I merge.

xuedong:
Hi, a small comment here just to mark a potential todo in the future: in some cases, there could be a tie breaking issue when choosing the next action/arm (if we simply use np.argmax). This is not a big problem in "real" data, but only for some synthetic experiments. Not urgent though.

Issue #151, opened 2022-03-10
Bug in plot_writer_data


omardrwch:
The following code:

```python
from rlberry.agents import UCBVIAgent
from rlberry.envs import GridWorld
from rlberry.manager import AgentManager, plot_writer_data

manager = AgentManager(
    UCBVIAgent,
    (GridWorld, {}),
    fit_budget=10,
    output_dir='rlberry_data/ucbvi_experiment',
    outdir_id_style=None
)
manager.fit()
plot_writer_data(manager, tag='episode_rewards')
```

raises the error:

```
ValueError: input dir not found, verify that the agent are trained and that AgentManager.outdir_id_style="timestamp"
```

Since the agent manager is given to `plot_writer_data`, we should not care about `outdir_id_style`, all we need is already in the manager.

Also, this leads to a completely unexpected behavior if the code above is previously called with `outdir_id_style="timestamp"`, since we'll plot the data from the previous run, instead of data in the current manager.

TimotheeMathieu:
Yes I agree. 
We should have some hierarchy on how to choose the files to plot. If the agentmanager is fitted -> plot these data. If it is not fitted but there are timestamp -> plot last data, if there is no timestamp and is not fitted, warn. 

Or maybe do something interactive in case not fitted and not timestamped, where plot_writer_data asks to select one choice from a list of possible folders ? what do you think ?

omardrwch:
I think that `plot_writer_data(input_obj, ..,)` should behave as follows:

* If `input_obj` is an AgentManager instance (or a list of instances), nothing should be loaded from memory: we plot the data stored in that instance
* If `input_obj` is a string/Path, we load the most recent manager from that path.

TimotheeMathieu:
Ok but If input_obj is a string/Path, we need the timestamps in order to be able to load most recent . 
What do we do if there is no timestamp ?
And also I think it would be nice to select not necessarily the most recent but an older example. Either interactively or by giving the exact path as argument.

omardrwch:
I think we could distinguish two cases when we have a path:

* If `input_obj` is a path to a .pickle file, we load the file as an AgentManager (which should allow us to load older experiments);
* If `input_obj` is a folder, we take the .pickle file in the most recent subfolder (either by looking at the timestamp in the folder name, or by looking at its creation time).

What do you think?

TimotheeMathieu:
Ok for me. I can do it.

PR #152, opened 2022-03-11
(fix) Solve issue plot_writer_data and handle file as argument

TimotheeMathieu:
Should solve Issue  #151 

read_writer_data and plot_writer_data now take as argument 
```
    input_obj : AgentManager, or list of AgentManager or directory or list of directories
        If AgentManager or list of AgentManager, load data from it (the agents must be fitted).
        If directory, load the data from the directory of the latest experiment in date.
        If list of directories, load the data from these directories.
```

Now these functions take as input 

- a fitted agent and in this case the data are plotted from memory, the functions will not use the pickle files.
- a directory containing subdirectories containing pickles in which case it will plot the pickles corresponding to the last experiment
- a list of directories containing pickles, in which case it will plot the pickles.

Question @omardrwch : I think we may want to rethink the default `rlberry_data` name of the directory. Would it make sense for example to use `rlberry_data/name_of_script/` instead ? The idea is that one directory should be for one script (= one experiment)

### DIFF
--- a/rlberry/manager/evaluation.py
+++ b/rlberry/manager/evaluation.py
@@ -135,13 +135,14 @@ def read_writer_data(input_obj, tag, preprocess_func=None):
         else:
             take_last_date = True
     else:
-        take_last_date = False
-        for dir in input_obj:
-            files = list(Path(dir).iterdir())
-            if len(files) == 0:
-                raise RuntimeError(
-                    "One of the files in input_obj does not contain pickle files"
-                )
+        if not isinstance(input_obj[0], AgentManager):
+            take_last_date = False
+            for dir in input_obj:
+                files = list(Path(dir).iterdir())
+                if len(files) == 0:
+                    raise RuntimeError(
+                        "One of the files in input_obj does not contain pickle files"
+                    )
 
     if isinstance(input_obj[0], AgentManager):
         agent_manager_list = input_obj
--- a/rlberry/manager/evaluation.py
+++ b/rlberry/manager/evaluation.py
@@ -1,13 +1,12 @@
 import logging
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


omardrwch:
I think it's ok to let only `rlberry_data` as default (or any other fixed string), and people can use the `output_dir` argument of `AgentManager` to name their experiments if they want.

But I think that another related thing can be useful: we could save a text file in the AgentManager's output_dir containing information about:
* Name of the script that launched the experiment (and maybe a hash of its contents);
* Timestamps (and runtime?) of each call to AgentManager.fit()
* Optimized hyperparameters (if any)
* etc, any other readable information that can be useful.

rlberry/manager/evaluation.py
```diff
@@ -96,19 +95,19 @@ def evaluate_agents(
...
 
-    This function can be called with agents that have been trained beforhand and
-    which have been saved in `agent_manager.output_dir`. The last session for each
-    individual agent will then  be loaded from the pickle files in `agent_manager.output_dir`.
-
     Parameters
     ----------
-    agent_manager : AgentManager, or list of AgentManager or directory
-        If AgentManager or list of AgentManager, load data from `agent_manager.output_dir`.
+    data_source : AgentManager, or list of AgentManager or directory or list of directories
```
omardrwch, line 105: When it can be a directory (or a list), what kind of directory are we expecting? Are those directories that we have as `output_dir` to `AgentManager`? Are they strings or Path objects? I think we should add this info to the documentation.

omardrwch (changes requested):
I think we can be more precise in the documentation for the directory/list of directories.

Also, I renamed `input_obj` to `data_source` (to make it clearer) and split the test into two cases (when `data_source` is an agent manager, and when it is a directory)

omardrwch (approved):

PR #159, opened 2022-04-08
feat (evaluation) add a warning to assess that enough MC simulations were used in evaluate_agent

TimotheeMathieu:
In this PR, I use a statistical test (Wilcoxon signed rank test) to assess that enough simulations were used in `evaluate_agent` to distinguish all the agents.

Example :
The following code of comparison between two reinforce agents
<details><summary>Python code</summary>

```python
import numpy as np
import matplotlib.pyplot as plt
from rlberry.agents.torch.dqn import DQNAgent
from rlberry.agents.torch import REINFORCEAgent
from rlberry.manager import AgentManager, read_writer_data, evaluate_agents
from scipy.stats import wilcoxon
from rlberry.envs.benchmarks.ball_exploration import PBall2D

env_ctor = PBall2D
env_kwargs={}
seed = 42
N_EPISODES = 10
GAMMA = 0.99
HORIZON = 50
BONUS_SCALE_FACTOR = 0.1
MIN_DIST = 0.1
params = {"gamma": GAMMA, "horizon": HORIZON,"learning_rate":0.05}
eval_kwargs = dict(eval_horizon=HORIZON, n_simulations=20)

if __name__ == "__main__":

    manager1 = AgentManager(
        REINFORCEAgent,
        (env_ctor, env_kwargs),
        fit_budget=N_EPISODES,
        init_kwargs=params,
        eval_kwargs=eval_kwargs,
        seed=seed,
        n_fit=10,
        parallelization="process",
    )
    manager1.fit()

    manager2 = AgentManager(
        REINFORCEAgent,
        (env_ctor, env_kwargs),
        fit_budget=N_EPISODES,
        init_kwargs={'gamma': 0.9, 'learning_rate': 0.8, "horizon":HORIZON},
        eval_kwargs=eval_kwargs,
        seed=seed,
        n_fit=10,
        agent_name="Reinforce2",
        parallelization="process",
    )
    manager2.fit()
    print("test over 5 simulations")
    evaluate_agents([manager1, manager2], n_simulations=5, plot=True)
    print('test over 20 simulations')
    evaluate_agents([manager1, manager2], n_simulations=20, plot=True)
```
</details>

gives the following message 

`[INFO] It is statistically difficult to differentiate between REINFORCE and Reinforce2. Either they have same evaluation or n_simulations was too small`

when n_simulations is 5 and no message otherwise.


KohlerHECTOR:
Hello, is it possible to also print the required min n_simulations for statistical significance ?

TimotheeMathieu:
Good question. 
I think we could give a non-conservative estimate under normality assumption. And we could give a very conservative estimate using empirical Chebychev inequality or something like that. I will think about this. 
Really the best solution would be sequential testing as discussed during the working group but this is hard.

mmcenta:
Hello Timothee!

I think this is a nice idea (and I learned about a new statistical test), but do you have an example of how to use it?

TimotheeMathieu:
New experiment: now I have a class `AgentComparer`, and with it I can compare two `AgentManager`.

For instance, for the example `comparison_agent.py` that I made, the output is like that:
```
[INFO] Reject the null after 2 groups 
[INFO] A2C_tuned is better than A2C 
```
which means that I used 2 groups of 8 fits (in all, 16 fit for each agent) instead of the 80 fits allowed for each agent !

Issue #160, opened 2022-04-08
A2C fix


KohlerHECTOR:
https://github.com/rlberry-py/rlberry/blob/8168dfc73a802dc9a6308e16a7d7bdce715d4f17/rlberry/agents/torch/a2c/a2c.py#L246-L273

It seems there is a useless for loop in the main training loop of A2C that is not in the original algorithm. After removing the loop, performances of the A2C agent match stable-baselines' A2C.  @riccardodv and I think the problem is that this loop induces repeated gradient steps in the same direction for k_epochs steps.

yfletberliac:
Good catch!
Also interesting to see that A2C does better without multiple gradient steps whereas PPO does.

mmcenta:
> Good catch! Also interesting to see that A2C does better without multiple gradient steps whereas PPO does.

It may be because PPO adds the constraint that the policy must be close to the one used for data collection. 

@KohlerHECTOR Any updates on this? I've seen on Slack that it might have been a hyperparameter issue?

KohlerHECTOR:
Yes I agree with @mmcenta , PPO does not have the same objective function as A2C. I think the bad performances of A2C were due to repeated gradient steps. There is another point that is that the rlberry A2C default hyperparameters are not the same as SB3's but it is not an issue.

PR #161, opened 2022-04-08
A2c matches sb3

KohlerHECTOR:
Hello, this a simple fix we did with @riccardodv for a2c (issue #160 ). We simply removed the for loop in question. We also include a test file that tests a2c on CartPole-v0, Acrobot-v1, PBall2D.


yfletberliac (approved):

TimotheeMathieu (commented):
LGTM, thanks ! 
My only comment is that we sadly cannot handle the tests you propose because they are too long for automatic checks on azure pipeline.  Maybe you can make this an example our of the tests this would be more efficient I think.

TimotheeMathieu:
Ok, all green, I merge.

PR #164, opened 2022-04-19
Improvements to AgentManager

omardrwch:
In this PR, I add some improvements to `Agent` and `AgentManager`:

* The default `env` and `train_env` are set to `None` (or `(constructor, kwargs) = (None, None)`), since not all algorithms require an environment. This may seem surprising, but we can think of problems beyond RL (as supervised learning, where rlberry can be very useful too) or, even in RL, we might have already some dataset from which we learn offline (instead of an environment). 

* In `AgentManager`,  the parameters in `init_kwargs_per_instance` are now used in `optimize_hyperparams`; and the `optimize_hyperparams` method now accepts a custom evaluation function that takes a list of trained agents as input, and outputs a scalar evaluating them.  Finally, the `AgentManager.eval_agents` method now accepts other `eval_kwargs` than the ones defined in the constructor, for more flexibility.


rlberry/manager/agent_manager.py
```diff
@@ -233,7 +235,7 @@ class AgentManager:
     def __init__(
         self,
         agent_class,
-        train_env,
+        train_env=(None, None),
```
xuedong, line 238: Minor: I see you have added typing somewhere else in the same file, maybe we can add them here as well? But it's ok, we can do a full pass afterward, not important.

rlberry/manager/agent_manager.py
```diff
@@ -470,21 +472,29 @@ def get_agent_instances(self):
             ]
         return []
 
-    def eval_agents(self, n_simulations: Optional[int] = None) -> list:
+    def eval_agents(
+        self,
+        n_simulations: Optional[int] = None,
+        eval_kwargs: Optional[dict] = None,
+    ) -> list:
```
xuedong, line 479: Another minor comment about the typing: I think it's probably better to precise also the underlying type of list, e.g. list[int] for [1, 2, 4]

rlberry/manager/agent_manager.py
```diff
@@ -1134,7 +1157,10 @@ def _optuna_objective(
         params_stats.fit()
 
         # Evaluate params
-        eval_value = np.mean(params_stats.eval_agents())
+        if not custom_eval_function:
```
xuedong, line 1160: Seems that custom_eval_function is not covered by tests?

omardrwch: fixed!

PR #165, opened 2022-04-22
[MRG] Improve RL Agent Checks

mmcenta:
Hello,

This PR improves the RL Agent Checks by *not* using the AgentManager when doing additive fits checks. 

As it turns out, using the agent manager can cause problems with the torch RNG - which this PR aims to solve by using only the Agent classes.

Additionally, I had to change a few things on the StableBaselinesAgent so it works with the new test, but it should be working well now :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ PyOpenGL_accelerate==3.1.5
 pyvirtualdisplay
 torch>=1.6.0
 stable-baselines3
+tensorboard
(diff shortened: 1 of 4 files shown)


mmcenta:
I think this should be finished, the remaining failing tests are caused by the `tensorboard` dependency.

mmcenta:
Merging because all tests are failing on the main branch.

Issue #166, opened 2022-04-22
Error when stable baselines is not installed


omardrwch:
Trying to use `rlberry.agents` raises an error if stable baselines is not installed due to this line:

https://github.com/rlberry-py/rlberry/blob/731879e986f155b13782ddc54c462515f374b774/rlberry/agents/__init__.py#L18

We could include a check before this line.

mmcenta:
Should we also produce a warning when it is not loaded?

TimotheeMathieu:
We can do a warning or we can avoid this by not putting it in the init. We say that people must import from `rlberry.agents.stable_baselines.SableBaselineAgent`. Kind of like when we say that we import from `rlberry.agents.torch.a2c` and not `rlberry.agents.a2c`.

mmcenta:
> We can do a warning or we can avoid this by not putting it in the init. We say that people must import from `rlberry.agents.stable_baselines.SableBaselineAgent`. Kind of like when we say that we import from `rlberry.agents.torch.a2c` and not `rlberry.agents.a2c`.

Ops, I made the PR before seeing this comment.

omardrwch:
I think requiring an import from rlberry.agents.stable_baselines.SableBaselineAgent is better than the warning. A warning in the init file would appear everytime we use any agent, even when we don't need the ones from stable baselines

mmcenta:
No problem, I think it's a fine solution too.

I can just delete the import from `__init__.py` (and the logger) in my PR if that's okay.

PR #167, opened 2022-04-22
[WIP] Make sure rlberry.agents works when StableBaselines3 is not installed

mmcenta:
As per #166, we introduced this bug with the StableBaselinesAgent.

I just added a guard that throws a warning if it can't find the library - let me know what you think.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -42,7 +42,7 @@ Basic Agents
    :toctree: generated/
    :template: class.rst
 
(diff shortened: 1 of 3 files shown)


TimotheeMathieu:
This should also solve the documentation problem (after appropriately changing the name in docs/api.rst).

omardrwch (approved):

TimotheeMathieu (approved):

Issue #172, opened 2022-05-02
PPO agent does not work (RunTimError: IndexError). 


KohlerHECTOR:
There is a mistake in the computations of the advantages and returns. Indeed, the `_compute_returns_avantages()` function assumes that list `rewards` is of size `self.horizon` . However it is not always the case since there is a `break` condition in the main loop of the `_run_episode()` function. The error can be seen when fitting a PPO agent on CartPole-v0 with default hyperparameters.

PR #173, opened 2022-05-02
Simple fix for issue #172

KohlerHECTOR:


TimotheeMathieu:
I tried your test on my computer and it works without your fix. Could you make a test that would have failed without your fix ? 
Also, when I increase the fit_budget in cartpole to 20 in your test I get
```
E           UnboundLocalError: local variable 'surr_loss' referenced before assignment
```
This is a separate bug I assume, no ? 
And maybe you could use `rlberry.envs.gym_make` instead of gym's make, I think this would be a bit cleaner.

KohlerHECTOR:
Hello, regarding the RunTimeError, this for me does not work withou the fix: 
`
from rlberry.agents.torch import PPOAgent

env = gym_make("CartPole-v0")
ppo = PPOAgent(
    **dict(
        env=env))
ppo.fit(1e4)
`

`
IndexError: list index out of range
`

TimotheeMathieu:
You are right, for me its the same. Maybe change for 
```
from rlberry.agents.torch import PPOAgent

env = gym_make("CartPole-v0")
ppo = PPOAgent(
  **dict(
        env=env))
ppo.fit(4)
```
which also does not work but which should not take too much time once fixed. 
Or something similar. 
Feel free also to fix the seed so that you are sure the environment will always behave in this way.

Issue #155, opened 2022-03-14
Bug with q_net_constructor in DQN


omardrwch:
I believe an else
```python
else:
    q_net_ctor = q_net_constructor
```
is missing after

https://github.com/rlberry-py/rlberry/blob/8168dfc73a802dc9a6308e16a7d7bdce715d4f17/rlberry/agents/torch/dqn/dqn.py#L153

(creating this issue so that I can double-check later)

xuedong:
hein ?

PR #179, opened 2022-05-04
DQN accepts user defined NN constructor

AleShi94:
Resolving issue https://github.com/rlberry-py/rlberry/issues/155#issue-1168922814


TimotheeMathieu (approved):
Maybe you can also change the docstring to say that `q_net_constructor` can also be a string in which case it is loaded via torch, and it can be None. Otherwise seems ok to me.

TimotheeMathieu:
You are right, it is not loaded with torch. From what I see, the string must be something like `my_awesome_lib.constructor` or something like that. It would be nice to explain this a bit more. @omardrwch would you have a one-liner that would explain this to put in the docstring ?

mmcenta (approved):
LGTM! The feature is useful and the docs are clear.

PR #180, opened 2022-05-05
[MRG] Allows to build NNs in A2C from a string equal to the function's name

riccardodv:
Allows to build value_net_fn and policy_net_fn from a string using rlberry.utils.factory.load. This is useful when you want to 
call model_factory_from_env to build the architectures from a .yaml file according to args specified therein


TimotheeMathieu:
It is not a big problem it is just that in the test for A2C, policy_net_fn  was always the default I guess so it does not test policy_net_fn which is string or callable. 
It would be nice to include a small test on this, maybe doing one iteration on some environment to check that there is no Runtime error. You could do this with @pytest.mark.parametrize for instance, it is clean (example of use in https://github.com/rlberry-py/rlberry/blob/85176dea56f4b2fe75d0e9e5374110e9f68c9462/rlberry/tests/test_agent.py#L53).

riccardodv:
@TimotheeMathieu I added some checks and is fine now

PR #185, opened 2022-05-09
[MRG] Improve A2C implementation

omardrwch:
Given all the issues with the current A2C implementation, I'm proposing the following modifications:

* Remove bonuses
* Remove reward normalisation
* Remove `horizon` as parameter: it's not necessary
* `batch_size` and `budget` are now measured in number of timesteps (instead of episodes)
* Replace `rlberry.agents.utils.memories.Memory`  by  `rlberry.agents.utils.replay.ReplayBuffer`. The code in `rlberry.agents.utils.memories` should not be used anymore, everything can be done by the new `ReplayBuffer` (on-policy algorithms as `A2C` only keep most recent data in the replay).

I didn't benchmark too much after the modifications, but I checked that it can solve CartPole:

```python
from rlberry.agents.torch import A2CAgent
import numpy as np

env = gym_make("CartPole-v0")
agent = A2CAgent(env, eval_interval=500)
agent.reseed(123)
agent.fit(50_000)

print(np.mean([agent.eval() for _ in range(5)]))
```


mmcenta (approved):
The only thing that kind of worries me is the use of the ReplayBuffer instead of a memory. If I understand correctly, this is fine because:
1) the ReplayBuffer has size `batch_size`;
2) even though the buffer starts back at the beginning when it runs out of space, this is fine since it only happens after a batch update.

Let me know if this is correct! Aside from that, LGTM.

riccardodv:
@mmcenta I agree with what you wrote.

riccardodv:
1. I think there are two functions doing the same thing basically. I am referring to 
https://github.com/omardrwch/rlberry/blob/e337ccfe4eb87ca7d3e3a329ef6e0aabebba3b2e/rlberry/agents/torch/a2c/a2c.py#L162-L168
and
https://github.com/omardrwch/rlberry/blob/e337ccfe4eb87ca7d3e3a329ef6e0aabebba3b2e/rlberry/agents/torch/a2c/a2c.py#L241-L245
I would just use maybe `policy(self, observation)` which is anyway requested in the implementation of the agent. It is not very clear to me why we use `_select_action(self, state)`, also in the previous implementation
2. As we were discussing today, I would also like to add some other loggs such as the KL between `cat_policy` and `cat_policy_old` and the average entropy of `cat_policy`. Do you think I can add this to your PR directly @omardrwch? Otherwise, I can confirm your changes and open a new one. In any case, I would like to ask you where is better to put some useful function like the one I use to compute KL in a stable way to avoid some division by zero problems that we had with torch implementation. Maybe in ../torch/utils/ or ../torch/a2c/ in the same file (a2c.py) or in another, and in case under which file's name?

riccardodv:
3. I am not sure the current implementation does really need to define a `cat_policy_old`. Am I correct? `cat_pilicy` is simply updated every `self.batch_size` and you don't need both policies at the same time in the code I have the feeling.

KohlerHECTOR:
> 1. I think there are two functions doing the same thing basically. I am referring to
>    https://github.com/omardrwch/rlberry/blob/e337ccfe4eb87ca7d3e3a329ef6e0aabebba3b2e/rlberry/agents/torch/a2c/a2c.py#L162-L168
>    and
>    https://github.com/omardrwch/rlberry/blob/e337ccfe4eb87ca7d3e3a329ef6e0aabebba3b2e/rlberry/agents/torch/a2c/a2c.py#L241-L245
>    I would just use maybe `policy(self, observation)` which is anyway requested in the implementation of the agent. It is not very clear to me why we use `_select_action(self, state)`, also in the previous implementation
> 2. As we were discussing today, I would also like to add some other loggs such as the KL between `cat_policy` and `cat_policy_old` and the average entropy of `cat_policy`, which I think could be useful to monitor the agents (what's your opinion also @omardrwch?). Do you think I can add this to your PR directly @omardrwch? Otherwise, I can confirm your changes and open a new one. In any case, I would like to ask you where is better to put some useful functions (e.g. the one I use to compute KL in a stable way to avoid some division by zero problems that we had with torch implementation of KL). Maybe in ../torch/utils/ or ../torch/a2c/ in the same file (a2c.py) or in another, and in case under which file's name?

Hello, I think the 2 functions are doing the same. You are right. However they are not called at the same time. policy() is used in evaluation mode. It means that it used to get an action with the trained policy. It is important to keep two distinct funcitons bc in some cases one wants to use a deterministic policy() in evaluation mode. For me, the most common thing I see in deep agents in other codes is a common function for both eval and training: 
```python
select_action(self, state, deterministic = False, eval = False):
  if eval: 
    with torch.no_grad():
      act_dist = self.cat_policy(state)
      if deterministic:
        a = argmax(act_dist) # mean if action distribution is normal
      else:
        a = act_dist.sample()
   else:
     act_dist = self.cat_policy(state)
     a = act_dist.sample()
   return a   
```

riccardodv:
I see @KohlerHECTOR, thank you very much.

riccardodv (approved):
Maybe in the future, we should check if `cat_policy` and `cat_policy_old` are both needed.

KohlerHECTOR:
Ok i will merge, hopefully I dont break anything.

PR #192, opened 2022-05-16
Experimental module

TimotheeMathieu:
Introduce an experimental module containing jax agents and ppo, avec, sac.


KohlerHECTOR:
LGTM. So you only changed the directory's organisation ?

KohlerHECTOR (approved):
Approved !

PR #202, opened 2022-05-31
feat (bandits) update tracker function and UCBV example

TimotheeMathieu:
Add the possibility to update the tracker with a function given as argument. Use this to code UCBV.


sauxpa (approved):
Approved

sauxpa (approved):
Even better!
Go

PR #204, opened 2022-06-01
Minor polish to UCB-V

sauxpa:


TimotheeMathieu (approved):
LGTM. Thanks

TimotheeMathieu:
Ah, Sorry, in fact I had something to say. Maybe with parameter c=0.1 it would work better ?

sauxpa:
c=0.1 might be better, but isn't more appropriate to keep the default value in the paper? I think if we want to honestly showcase the interest of UCB-V, we should test on other distrib e.g Beta with smaller variance so that empirical Bernstein can really shine.
(TBH from a bandit perspective: I don't think UCB-V works well in practice compared to more modern algo...)

TimotheeMathieu (approved):

PR #223, opened 2022-06-28
Ppobranch

AleShi94:
Tested PPO and compared to SB3


KohlerHECTOR:
LGTM !

KohlerHECTOR (approved):

mmcenta:
There are some changes I would like to make, but I think it may be best if we approve this and then I send a new PR.

Mainly I was refactoring the code so it's more like the DQN that Omar coded, adding a RolloutBuffer and some small implementation details for performance.

rlberry/agents/torch/ppo/ppo.py
```diff
@@ -98,48 +101,49 @@ def __init__(
...
+        self.learning_rate = learning_rate
+        self.eps_clip = eps_clip
+        self.k_epochs = k_epochs
         self.use_gae = use_gae
         self.gae_lambda = gae_lambda
+        # self.env = env
+        self.optimizer_type = optimizer_type
+        self.kwargs = kwargs
+
+        self.normalize_advantages = True  # TODO: turn into argument
```
TimotheeMathieu, line 123: Maybe do this todo ?

KohlerHECTOR:
> > There are some changes I would like to make, but I think it may be best if we approve this and then I send a new PR.
> > Mainly I was refactoring the code so it's more like the DQN that Omar coded, adding a RolloutBuffer and some small implementation details for performance.
> 
> What is RolloutBuffer? Otherwise, this version is using ReplayBuffer available in rlberry.agents.utils.replay, which is also used in the current versions of A2C and DQN. This PPO implementation should be close to the one of A2C.

A rollout buffer is a data structure (a class like a replay buffer) to store the transitions during an on policy algorithm iteration. It is the on policy version of the replay buffer

riccardodv (approved):

rlberry/agents/torch/ppo/ppo.py
```diff
@@ -207,110 +215,94 @@ def fit(self, budget: int, **kwargs):
             enconters a terminal state in which case it stops early.
```
TimotheeMathieu, line 215: This description of the budget is not true anymore, is it ? Isn't the budget the number of steps ?

TimotheeMathieu (commented):
LGTM, thanks. 

Some minor comments:

- Long test implementation ?
- You can do a `git grep PPOAgent` to see where are the import for PPO that you still need to change. Typically, in api.rst and in the examples.
- Can you add this to the Changelog as this is an important change that deserves to be in the changelog.

riccardodv:
> > > There are some changes I would like to make, but I think it may be best if we approve this and then I send a new PR.
> > > Mainly I was refactoring the code so it's more like the DQN that Omar coded, adding a RolloutBuffer and some small implementation details for performance.
> > 
> > 
> > What is RolloutBuffer? Otherwise, this version is using ReplayBuffer available in rlberry.agents.utils.replay, which is also used in the current versions of A2C and DQN. This PPO implementation should be close to the one of A2C.
> 
> A rollout buffer is a data structure (a class like a replay buffer) to store the transitions during an on policy algorithm iteration. It is the on policy version of the replay buffer

But if I understand correctly we are using replay buffer also for on policy algorithms (like in DQN), we are just careful updating the policy every time we reach the total capacity of the buffer, right? (in this way we stay on policy)

KohlerHECTOR:
> a replay buffer that is emptied in between each policy updates is equivalent to a rollout buffer.

exactl

> > > > There are some changes I would like to make, but I think it may be best if we approve this and then I send a new PR.
> > > > Mainly I was refactoring the code so it's more like the DQN that Omar coded, adding a RolloutBuffer and some small implementation details for performance.
> > > 
> > > 
> > > What is RolloutBuffer? Otherwise, this version is using ReplayBuffer available in rlberry.agents.utils.replay, which is also used in the current versions of A2C and DQN. This PPO implementation should be close to the one of A2C.
> > 
> > 
> > A rollout buffer is a data structure (a class like a replay buffer) to store the transitions during an on policy algorithm iteration. It is the on policy version of the replay buffer
> 
> But if I understand correctly we are using replay buffer also for on policy algorithms (like in DQN), we are just careful updating the policy every time we reach the total capacity of the buffer, right? (in this way we stay on policy)

exactly ! a replay buffer that is emptied in between each policy updates is equivalent to a rollout buffer.

PR #226, opened 2022-07-13
feat (logging) Improve logging

TimotheeMathieu:
In this PR I make the following changes:

- There is now one logger `rlberry.logger`. It is used everywhere instead of `logging.getLogger(__name__)`. This allows us to handle logging more easily in child processes when doing multiprocessing. This fix the bug that the level could not be changed before.
- The level is fixed usind `rlberry.utils.logging.set_level`.
- The level of worker process in `AgentManager` can still be changes and it can be different from the main logger level.
- I changed the level of some of the messages to be debug
- I added a kind of progress bar for evaluation when level is info (one point = one evaluation)
- I made it so that when we are at info level, gym is at warning. And when we are at debug, gym is at info. And when we are at notset (minimal level), gym is at debug (gym's minimal level)

Example:

<details><summary>Python code</summary>

```python
from rlberry.utils.logging import set_level
from rlberry.agents.torch import A2CAgent
from rlberry.manager import AgentManager, evaluate_agents
import numpy as np
from rlberry.agents.torch.utils.training import model_factory_from_env
import numpy as np

set_level('DEBUG')

# Using parameters from deeprl quick start
policy_configs = {
    "type": "MultiLayerPerceptron",  # A network architecture
    "layer_sizes": (64, 64),  # Network dimensions
    "reshape": False,
    "is_policy": True,
}

critic_configs = {
    "type": "MultiLayerPerceptron",
    "layer_sizes": (64, 64),
    "reshape": False,
    "out_size": 1,
}
n_steps = 5e3
batch_size= 128


if __name__=="__main__":

    manager = AgentManager(
        A2CAgent,
        (gym_make, dict(id="CartPole-v1")),
        agent_name="A2CAgent",
        init_kwargs=dict(
            policy_net_fn=model_factory_from_env,
            policy_net_kwargs=policy_configs,
            value_net_fn=model_factory_from_env,
            value_net_kwargs=critic_configs,
            entr_coef=0.0,
            batch_size=1024,
            optimizer_type="ADAM",
            learning_rate=1e-3,
        ),
        fit_budget=3e4,
        eval_kwargs=dict(eval_horizon=500),
        n_fit=2,
        parallelization="process",
        mp_context="spawn",
        seed=42,
         )
    manager.fit()
    outputs = evaluate_agents([manager], n_simulations=20, choose_random_agents=False, show=False)
    manager.fit()
```

</details>

Results when in DEBUG level :
debug

Results when INFO level:
info2

Any comment are welcome.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -176,6 +176,15 @@ Check Utilities
    utils.check_seeding_agent
    utils.check_agent_manager
 
+Logging Utilities
+-----------------
+
+.. autosummary::
+   :toctree: generated/
+   :template: function.rst
+
+   utils.logging.set_level
+
 
 Typing
 ------
--- a/examples/demo_bandits/plot_mirror_bandit.py
+++ b/examples/demo_bandits/plot_mirror_bandit.py
@@ -24,7 +24,9 @@ import requests
 import matplotlib.pyplot as plt
 
 
-logger = logging.getLogger(__name__)
+import rlberry
+
+logger = rlberry.logger
 
 # Environment definition
 
--- a/examples/demo_bandits/plot_mirror_bandit.py
+++ b/examples/demo_bandits/plot_mirror_bandit.py
@@ -19,7 +19,7 @@ from rlberry.envs.interface import Model
 from rlberry.agents.bandits import BanditWithSimplePolicy
 from rlberry.wrappers import WriterWrapper
 import rlberry.spaces as spaces
-import logging
+
 import requests
 import matplotlib.pyplot as plt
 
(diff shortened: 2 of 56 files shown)


TimotheeMathieu:
I added some colors ! (wow)
image

the errors (that we don't see on the screenshot) are in red.

KohlerHECTOR (approved):

TimotheeMathieu:
I introduce a new style for the writer logger called `multi_line`, the old style is now called `one_line`. 

Here is a preview:
image

Should this be the default (with fallback to one_line if size of terminal window is too small) ?

AleShi94 (approved):

AleShi94 (approved):

PR #227, opened 2022-07-20
feat (plot) add smoothing, tensorboard style

TimotheeMathieu:
This PR make it so that the plot is smooth in `plot_writer_data` when there have been only one fit.

Example of result : 
image

There is a parameter `smooth_weight` in `plot_writer_data` to regulate the amount of smoothing.

### DIFF
--- a/rlberry/manager/evaluation.py
+++ b/rlberry/manager/evaluation.py
@@ -295,7 +295,8 @@ def plot_writer_data(
     title=None,
     savefig_fname=None,
     sns_kwargs=None,
-    smooth_weight=0.0,
+    smooth_weight=0.09,
+    plot_raw_curves=True,
 ):
     """
     Given a list of AgentManager or a folder, plot data (corresponding to info) obtained in each episode.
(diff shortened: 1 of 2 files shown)


KohlerHECTOR:
Very nice feature ! Maybe for safety it would be nice to have in the legend "dark blue: smoothed curve ; light blue: non-smoothed curve ".

KohlerHECTOR:
> Proposition:
> 
> image
> 
> I show two agents here to show that this enable us to do better plot when axis is dw_time_elapsed.

Hello, why is it better plot to use elapsed time in x axis ?

KohlerHECTOR:
> What I wanted to say is that, if you choose to use dw_elapsed_time as an x axis (which impose that we use only one simulation). Smoothing things give you a visually better plot.

nice !

AleShi94 (approved):
Very nice!

TimotheeMathieu:
Ok, thanks for the reviews. 
We may want to have a few examples using this feature, to plot reward of rl agent on one run for instance. This may be done in a separate PR, for now I merge.

PR #254, opened 2022-11-16
Qlearning and Sarsa

KohlerHECTOR:
## Description

Very early pull request for QL and SARSA. 

## TODO

Add docs.


JulienT01:
Some idea:


Everywhere : 
Change the variable names 'a / s / r'   to 'action / observation or state / reward' to be consistent with other modules (and avoid one letter variable)

In test_tabular_rl:
- use a bigger value for the "fit" (to trigger the reset function)
- test the "policy" function
- test the sarsa and qlearning with "tau" too, not only "eps-greedy"

For doc:
- add 1 example for qlearning and 1 for sarsa. It could be nice for the docs  (you can check some examples in /examples/demo_agent/video_plot_*.py)

PR #296, opened 2023-03-17
xfail on failing tests with mac and windows

JulienT01:
## Description

Add the decorator "xfail" to pass the failing tests with Mac and Windows

### DIFF
--- a/rlberry/manager/tests/test_agent_manager.py
+++ b/rlberry/manager/tests/test_agent_manager.py
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+import sys
 import os
 from rlberry.envs import GridWorld
 from rlberry.agents import AgentWithSimplePolicy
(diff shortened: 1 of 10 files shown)


JulienT01:
Working with MAC, but there is some problemes with windows:
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'rlberry_data\\temp\\manager_data\\_2023-03-17_09-13-38_a3880128\\optuna_data.db'

https://dev.azure.com/timotheemathieu/timotheemathieu/_build/results?buildId=1046&view=logs&j=3c0a9d00-a287-5cb5-e420-b7a0de833172&t=76e2acb6-c9a2-5eb0-53d2-31ed6586c55e&l=1870

rlberry/manager/tests/test_agent_manager.py
```diff
@@ -165,6 +169,7 @@ def test_agent_manager_2():
     # test hyperparemeter optimization
     loaded_stats.optimize_hyperparams(n_trials=5)
 
+    time.sleep(3)
```
TimotheeMathieu, line 172: This can be removed

rlberry/tests/test_agent.py
```diff
@@ -57,6 +58,7 @@ def test_finite_state_agent(agent):
     check_rlberry_agent(agent, env="discrete_state")
 
 
+@pytest.mark.xfail(sys.platform == "win32", reason="bug with windows???")
```
TimotheeMathieu, line 61: Maybe we can do another test for windows specifically, in which we don't test PPO. With this xfail, you are xfailing the test for all agents even though only one agent fails (I think)

PR #299, opened 2023-03-31
PPO buffer crash with multi-fit

JulienT01:
## Description

With the previous version, the buffer for PPO could be in "Buffer is full." when it was not a single fit().
This PR add some tests to cover this specificity, and a correction to this bug.

## Checklist
- \[x] I have commented my code, particularly in hard-to-understand areas
- \[x] I have made corresponding changes to the documentation
- \[x] I have added tests that prove my fix is effective or that my feature works
- \[x] New and existing unit tests pass locally with my changes
- \[x] I have set the label "ready for review" and the checks are all green


rlberry/agents/torch/tests/test_ppo.py
```diff
@@ -197,5 +201,363 @@ def test_ppo():
...
+def test_ppo_multi_env():
+    env = gym_make("CartPole-v0")
+    agent = PPOAgent(env, learning_rate=1e-4, optimizer_type="ADAM", n_envs=3)
+    agent.fit(budget=1000)
+
+    saving_path = "rlberry/agents/torch/tests/agent_test_ppo_classic_env.pickle"
+
+    # VRemove previous save
+    if os.path.exists(saving_path):
+        os.remove(saving_path)
```
TimotheeMathieu, line 250: Another way to do this is to use temporary folder. Python has a function for this. It is convenient because : unique name, and deleted automatically at reboot. In particular, this is convenient to not have litter files when running test locally. I think I made some of the rlberry tests with temporary folder (check out here: https://github.com/rlberry-py/rlberry/blob/7dfba03559258dfe512bee231f21ab95bb9db2af/rlberry/utils/check_agent.py#L181)

This comment is also applicable to the rest of the tests.

rlberry/utils/check_agent.py
```diff
@@ -222,6 +250,95 @@ def check_seeding_agent(agent, env=None, continuous_state=False, init_kwargs=Non
...
+
+    Parameters
+    ----------
+    agent: rlberry agent module
+        Agent class to test.
+    env: tuple (env_ctor, env_kwargs) or str in {"continuous_state", "discrete_state","vectorized"}, default="vectorized_env_continuous"
+        if tuple, env is the constructor and keywords of the env on which to test.
+        if str in {"continuous_state", "discrete_state","vectorized"}, we use a default Benchmark environment.
+    agent_init_kwargs : dict
+        Arguments required by the agent's constructor.
```
TimotheeMathieu, line 309: I think you don't need env that can be continuous_state or discrete_state. It can only be vectorized_env_continuous and maybe a tuple.

TimotheeMathieu (approved):
A small comment on doc, but otherwise LGTM.

PR #312, opened 2023-05-30
[minor changes] New CI,  "Optuna" in Check-agent, and "profile.prof" .gitignore

JulienT01:
PR to add some minor changes

- Add profile.prof into the .gitignore. (Artifact created by testing agent_manager)
- Move the "optuna" import inside the function that needs it, to avoid crashes on other functions (when you don't install optuna).
- Add a CI to test the installation without requirement.txt and without the editable "-e" mode
- Remove the 'network' folder from the code coverage

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -135,3 +135,60 @@ jobs:
       pip install pytest==7.0.1 pytest-azurepipelines pytest-xvfb
       pytest --ignore=rlberry/network
     displayName: 'pytest'
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+- job: 'linux_non-editable'
+  dependsOn: checkPrLabel
+  # Condition: have ready for review label or on the main branch.
+  condition: or(in(variables['Build.SourceBranch'], 'refs/heads/main'), eq(dependencies.checkPrLabel.outputs['checkPrLabel.prHasCILabel'], true))
+  pool:
+    vmImage: ubuntu-latest
+  strategy:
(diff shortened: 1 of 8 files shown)


azure-pipelines.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       # run doctests in the documentation
       # TODO: use following line for doctest ?
       pytest docs/*rst || echo "Ignoring exit status"
-      pytest --import-mode=importlib --cov=rlberry --cov-report xml rlberry
+      pytest --import-mode=importlib --cov=rlberry --cov-report xml rlberry --ignore=rlberry/tests/test_no_extra_install.py
```
TimotheeMathieu, line 67: Maybe do not ignore it here and then it will figure in the coverage report.

TimotheeMathieu (approved):
LGTM. Maybe in this PR you can also ignore the network folder for codecov ? (see https://github.com/rlberry-py/rlberry/blob/5bbe27a1345b0535a00e57734b8fb1728f1c22f8/codecov.yml#L23)

azure-pipelines.yml
JulienT01: > Maybe do not ignore it here and then it will figure in the coverage report.

I had ignored it because it was duplicate testing of "base" agents and envs.
Following your message, to add it to the code coverage, I reorganized the agent tests into "base" and "extra", which removes the duplicates and allows you to test specifically the ones you want (the code coverage checks everything, and the CI "base" check only the "base" agents and envs).

PR #318, opened 2023-06-09
Load cpu gpu agent manager

JulienT01:
Update Code to be able the saves/loads cross CPU/GPU.  (train and save an agent on GPU and load it on computer with only CPU)

- add a class "AgentTorch" (inherited from Agent) for torch Agent :
    - Overwrite the save/load functions with torch.save / torch.load
    - update the agents using torch to be inherited from "AgentTorch"
- add an option to choose to save or not the environment with the agent
- update a test in check agent that was trying to load an agent with agentManager

## Checklist
To check :
- \[x] I have made corresponding changes to the documentation,
- \[ ] I have added tests that prove my fix is effective or that my feature works,


TimotheeMathieu (approved):
LGTM. Maybe you could just make it so that there is an error message/warning when trying to load idx0.pickle in an agent manager (I mean, any pickle which is not manager_obj.pickle). Otherwise All is ok for me.

PR #326, opened 2023-06-29
Add Soft Actor-Critic (SAC) 

brahimdriss:
## Description

This PR introduces SAC, following the original article, for continuous action spaces.
The current implementation was evaluated on gym Pendulum and Mujoco Hopper (v2), "solving" both environments.

## Todo
- \[x] Switch to gymnasium
- \[x] Add seeding
- \[x] Handle multiple fitting
- \[x] Add benchmarks (Mujoco and classic gym control)

**Original articles:**
[Soft Actor-Critic: Off-Policy Maximum Entropy Deep Reinforcement Learning with a Stochastic Actor](https://arxiv.org/abs/1801.01290)
[Composable Deep Reinforcement Learning for Robotic Manipulation](https://arxiv.org/abs/1803.06773)
[Soft Actor-Critic Algorithms and Applications](https://arxiv.org/abs/1812.05905)

**Reference resources:**
[haarnoja/sac](https://github.com/haarnoja/sac)
[haarnoja/softqlearning](https://github.com/haarnoja/softqlearning)
[openai/spinningup](https://github.com/openai/spinningup/tree/master/spinup/algos/tf1/sac)
[vwxyzjn/cleanrl](https://github.com/vwxyzjn/cleanrl)
[DLR-RM/stable-baselines3](https://github.com/DLR-RM/stable-baselines3/tree/master/stable_baselines3/sac)

## Checklist
To check :
- \[x] I have made corresponding changes to the documentation,
- \[x] I have added tests that prove my fix is effective or that my feature works,


brahimdriss:
Pendulum results on 10 runs. Not sure about the style of the plot, but it still can be change later.

sac_pendulum

MuJoCo results are next.

brahimdriss:
### **MuJoCo results:**

**Hopper-v2 :** 1M timesteps - 5 runs 
Training episodic return : **2247.09 ± 929.56**
sac_hopper_v2

rlberry/agents/torch/sac/sac.py
```diff
@@ -0,0 +1,543 @@
...
+
+        # initialize episode, steps and time counters
+        self.total_episodes = 0
+        self.total_timesteps = 0
+        self.time = time.time()
+
+    def policy(self, state):
+        assert self.cont_policy is not None
+        state = np.array([state])
+        state = torch.FloatTensor(state).to(self.device)
```
riiswa, line 228: Just a nit, but there is some built-in Pytorch method to convert a np array to a torch tensor (`torch.as_tensor`, `torch.from_numpy`). I think `torch.as_tensor` allow you to specify the device directly.

rlberry/agents/torch/sac/sac.py
```diff
@@ -0,0 +1,543 @@
...
+        )
+
+        with torch.no_grad():
+            # Select action using the current policy
+            next_state_actions, next_state_log_pi = self._select_action(
+                next_state.detach().cpu().numpy()
+            )
+            # Compute the next state's Q-values
+            q1_next_target = self.q1_target(
+                torch.cat([next_state, next_state_actions], dim=1)
```
riiswa, line 402: Is this compatible with multidimensional shapes observations like images ? If not I suggest to use dim=-1

brahimdriss: I think you are right, I'll change it to `dim=-1`.

rlberry/agents/torch/sac/sac.py
brahimdriss: Thanks for the suggestions Waris,  I tried here to be consistent with the other implementations of Rlberry, `FloatTensor` was used in some of them, but so was `torch.from_numpy`. That's why I kept it. Is there any difference in terms of performance otherwise ?

KohlerHECTOR:
It seems that something is broken with the doc in the sac branch. When trying to change the api.rst , it breaks readthedocs compilation. So I reverted. But otherwise, algorithmically speaking, SAC looks good to me.

Issue #286, opened 2023-03-03
Bugs in Torch Model Factory


mmcenta:
Hello,

Yesterday, I noticed that the model factory functions we have for torch can have some problems when trying to use PPO with continuous actions. It should be a quick fix, but we also had these issues with creating neural networks with shared layers - so I think it might be best if we rework how these model factories work. I'll think about it and open a PR later about this, but you can let me know if there are any problems/feature requests for the model factory utils for torch.

KohlerHECTOR:
This issue is related to @Philippe's request to have utilities to manipulate actor, critics, and other networks. In particular to do transfer learning on DQN. So we plan to rework the model factory. 
Ideas are welcome in advance of the rlberry hackathon @mmcenta @AleShi94 @omardrwch @riccardodv @brahimdriss @yann .

PR #351, opened 2023-07-24
Model factory

AleShi94:
Possible solution to issue #286

1) `model_factory` and `model_factory_from_env` can take external neural network.
2) `model_factory` and `model_factory_from_env` can load a nn from file
3) `model_factory` and `model_factory_from_env` can load a checkpoint for nn
4) `model_factory_from_env` runs the check that the external nn satisfies requirements of the environment and model_configs

All new features are tested in `test_torch_training.py` file


KohlerHECTOR (commented):
I am not sure I understand the if/else structure of the model factory. What happens if both a filename and nn.module instance are passed to the model factory function ?

AleShi94:
> I am not sure I understand the if/else structure of the model factory. What happens if both a filename and nn.module instance are passed to the model factory function ?

well, technically you can do it so that you take architecture from nn.Module and from the file you can take model weights. In that case it will load weights from the file and assign them to the architecture

But to be honest I am not 100% sure if it is the best way to code stuff, but the idea is that it should be flexible enough:
either you use available tools to build a nn (like it was done before), or you can propose your own architecture through net argument or you can upload your model or checkpoint of the model through a file, which should allow transfer learning

KohlerHECTOR:
Your idea is very clear and cool. What about only taking a nn module instance as input ? And providing the user with a simple pol net val net and a model factory ? So it is the user's task to build its networks the way he/she likes?

JulienT01:
I just added two little comments, but otherwise it looks good to me. :+1:

PR #358, opened 2023-07-25
Add attributes to A2C

riccardodv:
Added attributes to A2C, check for the future if it is better to eliminate redundancy with parameters.


TimotheeMathieu:
I think it would be better to eliminate redundancy. If people want to get the parameters after instantiation, there is already `get_params` that is documented so they can use that without having to be said which parameters are set to attribute and which are not.

KohlerHECTOR:
Good to mergne when duplicates are removed :).

riccardodv:
I will do it soon, but I am on vacation these days, and I am travelling.

Issue #374, opened 2023-10-12
evaluate_agents do more evaluation than 'n_simulation' param


JulienT01:
evaluate_agents on ExperimentManager do more evaluation than the n_simulation  (for PPO)

the n_simulation parameter go to eval_agents() in ExperimentManager, but not to eval() in Agent :  kwarg parameter don't have the n_simulation anymore.
So the eval() is done 10 times (default value)

We need to add n_simulation=1 in the eval_kwargs for the line (experiment_manager.py)
"values.append(agent.eval(**eval_kwargs))"

BorisHamadej:
HI! Is this issue still open? Can I give it a try?

JulienT01:
Yes it's still open, you can do it if you want.

TimotheeMathieu:
Solved, closing.

Issue #369, opened 2023-10-03
Upgrade or remove Seaborn


JulienT01:
Since september 2013, seaborn upgraded to 0.13.0... Now we have 4 failing tests on "rlberry/manager/tests/test_agent_manager.py".

We need to update our code to make tests green (updtaing code to remain compatible with seaborn, or removing seaborn).

PR #375, opened 2023-10-12
[MRG] Added n_simulation in the eval_kwargs

BorisHamadej:
Created a new dictionary "eval_kwargs_with_n_simulatio"n that is a copy of eval_kwargs with the additional n_simulation parameter set to 1. Issue fixed is #374: https://github.com/rlberry-py/rlberry/issues/374.
My motivation was to contribute to this project, it was labeled good first issue.

### DIFF
--- a/rlberry/manager/experiment_manager.py
+++ b/rlberry/manager/experiment_manager.py
@@ -573,7 +573,11 @@ class ExperimentManager:
                     " Returning []."
                 )
                 return []
-            values.append(agent.eval(**eval_kwargs))
+            # Update eval_kwargs with n_simulation parameter
+            eval_kwargs_with_n_simulation = eval_kwargs.copy()
+            eval_kwargs_with_n_simulation["n_simulation"] = 1
+
+            values.append(agent.eval(**eval_kwargs_with_n_simulation))
             if verbose:
                 if logger.getEffectiveLevel() <= 10:  # If debug
                     logger.debug(f"[eval]... simulation {ii + 1}/{n_simulations}")
--- a/rlberry/manager/experiment_manager.py
+++ b/rlberry/manager/experiment_manager.py
@@ -30,7 +30,6 @@ from rlberry.utils.writers import DefaultWriter
 from rlberry.manager.utils import create_database
 from rlberry import types
 
-
 _OPTUNA_INSTALLED = True
 try:
     import optuna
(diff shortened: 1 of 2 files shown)


rlberry/manager/experiment_manager.py
```diff
@@ -573,7 +573,11 @@ def eval_agents(
                     " Returning []."
                 )
                 return []
-            values.append(agent.eval(**eval_kwargs))
+            # Update eval_kwargs with n_simulation parameter
+            eval_kwargs_with_n_simulation = eval_kwargs.copy()
+            eval_kwargs_with_n_simulation["n_simulation"] = 1
+
+            values.append(agent.eval(**eval_kwargs_with_n_simulation))
```
TimotheeMathieu, line 580: Thanks, LGTM! 
Maybe a small comment though: could you add a warning (with `logger.info`) so that when `n_simulation` already exists in  `eval_kwargs`, we have a warning saying that it is overwritten. 
eval_kwargs is not meant to be used like that but if people do it, let us give a warning that we are overwriting with 1.

BorisHamadej: I added the warning, thank you for the help. Is it looking good now?

TimotheeMathieu:
The test is failing because of a `DummyAgent` constructed for a test. Adding `**kwargs` to the arguments of the eval function in https://github.com/rlberry-py/rlberry/blob/d9378eb68ffd7d13686e6afe971b7835587d9e39/rlberry/manager/tests/test_comparisons.py#L26 should solve this.

rlberry/manager/experiment_manager.py
TimotheeMathieu: Thanks. There are still some tests that do not pass, see my comment at the end of the discussion and the "detail" button the checks at the bottom. Some test was broken by the change.

BorisHamadej: I added the argument but some tests are still failing

JulienT01:
Thank you for your contribution, LGTM!

rlberry/manager/experiment_manager.py
TimotheeMathieu: The tests are failing due to https://github.com/rlberry-py/rlberry/issues/369 , this is ok.

PR #376, opened 2023-10-20
[WIP] new feature: confidence band around reward curve and proper smoothing

TimotheeMathieu:
In this PR I introduce a function to plot the reward curves which 

1) smooth the curves properly using a Nadaraya-Watson estimator based on Gaussian windows functions whose bandwith can be either set manually or computed through cross-validation. 
2) plot either all the raw curves or plot a confidence band of the mean smoothed curve. The confidence interval is based on bootstrap hence it is non-parametric.
3) For an (almost) retro-compatibility, I also added the possibility to use "ci" which plot mean plus/minus quantile times std/sqrt(n_seed) and "pi" which plot mean plus/minus quantile times  std. The use of "pi" should be close to what we had before.

The confidence band is a confidence interval around the mean smoothed curve, i.e. one reward curve is counted as one random variable and we use functional data analysis to compute the mean and confidence interval of this function.

An added bonus is that I do everything in matplotlib hence it remove the need to have seaborn as dependency. On the other hand this add the dependency [scikit-fda](https://fda.readthedocs.io)  for the fda part of the feature.

Image (two calls to the new plot_writer_data: on the left plotted with raw curves on the right with the ci):

image
Illustration of the four different error representations now:
image

Refs:
- [scikit-fda](https://fda.readthedocs.io)
- The ci computation is based on [this article](https://wires.onlinelibrary.wiley.com/doi/full/10.1002/wics.1397) and [this one](https://www.jstor.org/stable/24309653).

Still todo: 
- [x] Be able to plot prediction intervals.
- [x] Be able to plot same prediction intervals as previously.
- [x] ~~Maybe do an example where we use seaborn lineplot instead (i.e. replicate the old behaviour with the new plot_writer_data function because plot_writer_data give the dataframe which can be directly use for plotting)~~ -> Will do in user guide
- [x] Add references
- [x] Add examples
- [x] ~~Add doc in user guide~~ -> will do later
- [x] Do tests
- [x] Update changelog


JulienT01 (commented):
I've added some minor comments, but it looks good to me.

JulienT01 (approved):

PR #380, opened 2023-11-06
Update Tabular RL agents to have epsilon-greedy exploration strategy by default with eps=0.3

riiswa:
## Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context, in particular link the relevant issue if it is appropriate.
List any dependencies that are required for this change.


riiswa:
Not needed anymore, Tabular RL agents will be added to rlberry-scool [#5](https://github.com/rlberry-py/rlberry-scool/pull/5)

PR #397, opened 2023-12-12
automatic save in ExperienceManager

JulienT01:
- Add a save at the end of the fit() of ExperienceManager
- Add a warning when there is a previous save with the same path
- Add a logger.info to display the path of the current save

## Checklist
To check :
- \[x] I have added tests that prove my fix is effective or that my feature works,

### DIFF
--- a/rlberry/manager/tests/test_experiment_manager.py
+++ b/rlberry/manager/tests/test_experiment_manager.py
@@ -438,7 +438,7 @@ def test_save_logger_and_warning(caplog):
         path = stats_agent1.save()
         assert str(ExperimentManager_message_begin + str(path) + "'") in caplog.text
 
-        warning_overwrite_message = "This output directory already exists, the save may change the symbolic link or overwrite the previous Experiment."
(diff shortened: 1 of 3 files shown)


rlberry/manager/tests/test_experiment_manager.py
```diff
@@ -403,3 +405,51 @@ def test_logs(style_log):
...
     evaluate_agents([stats], show=False)
+
+
+def test_save_logger_and_warning(caplog):
+    with tempfile.TemporaryDirectory() as tmpdir:
+        agent_test_name = "test_agent"
+
+        path_to_save = tmpdir + "/" + agent_test_name
+        if os.path.isdir(path_to_save):
+            shutil.rmtree(path_to_save)
```
TimotheeMathieu, line 416: Typically, this can be ignored by the coverage no?

JulienT01: yes , but in other PR https://github.com/rlberry-py/rlberry/pull/396

TimotheeMathieu (approved):

PR #404, opened 2024-02-13
add AdaStop

TimotheeMathieu:
This PR adds AdaStop algorithm to rlberry.

Typical example is what I added as an example, a very simple script that test A2C versus PPO from sb3 on Acrobot with default parameters. Results is:

```
[INFO] 13:35: Test finished 
[INFO] 13:35: Results are  
  Agent1 vs Agent2  mean Agent1  mean Agent2  mean diff  std Agent 1  std Agent 2 decisions
0       A2C vs PPO     -274.274      -85.068   -189.206    185.82553      2.71784   smaller
```

TODO:
- [x] Run the example with more seeds to differentiate which is better
- [ ] Add part in user guide about this (will do later)
- [x] Docstrings
- [x] tests


JulienT01 (approved):
LGTM

Issue #395, opened 2023-11-28
Benchmark downloader and processor


TimotheeMathieu:
It would be nice to have a function to fetch and process benchmark results from various places like stablebaselines3-zoo or cleanrl, so they can be used easily to compare with the results of a rlberry user.

TimotheeMathieu:
Google Atari bucket: https://console.cloud.google.com/storage/brow ser/rl-benchmark-data/ALE
SB3: https://github.com/DLR-RM/rl-baselines3-zoo/tree/master/logs/benchmark WARNING only one seed
cleanrl: https://wandb.ai/openrlbenchmark/openrlbenchmark/reportlist (no idea what data they have)

There are other.

Goal is to fit entry format for comparator (https://rlberry-py.github.io/rlberry/basics/comparison.html).

@YannBerthelot , @KohlerHECTOR, @mmcenta, @AleShi94  do you have any other idea of benchmark that we could use ?

JulienT01:
link to Google Atari bucket: https://console.cloud.google.com/storage/browser/rl-benchmark-data/atari_100k
associated colab : https://colab.research.google.com/drive/1a0pSD-1tWhMmeJeeoyZM1A-HCW3yf1xR?usp=sharing#scrollTo=eV8ci5frgPFO
associated paper : https://arxiv.org/pdf/2108.13264.pdf

TimotheeMathieu:
There is also https://github.com/DLR-RM/rl-trained-agents/

KohlerHECTOR:
La classe @YannBerthelot !!! Viens tu présentes aujourd'hui ?

PR #449, opened 2024-04-05
[WIP] Benchmark downloader

JulienT01:
Tools to download some benchmark.

linked to the issue : https://github.com/rlberry-py/rlberry/issues/395

## Checklist
- \[ ] My code follows the style guideline
To check :
- \[ ] I have commented my code, particularly in hard-to-understand areas,
- \[ ] I have made corresponding changes to the documentation,
- \[ ] I have added tests that prove my fix is effective or that my feature works,
- \[ ] New and existing unit tests pass locally with my changes,
- \[ ] If updated the changelog if necessary,
- \[ ] I have set the label "ready for review" and the checks are all green.

### DIFF
--- a/rlberry/manager/comparison.py
+++ b/rlberry/manager/comparison.py
@@ -27,11 +27,9 @@ def compare_agents(
     Parameters
     ----------
     agent_source : list of :class:`~rlberry.manager.ExperimentManager`, list of str
-
         - If list of ExperimentManager, load data from it (the agents must be fitted).
-
         - If str, each string must be the path of a agent_manager.obj.
-
+        **Each agent must have unique name.**
     method: str in {"tukey_hsd", "permutation"}, default="tukey_hsd"
         Method used in the test. "tukey_hsd" use scipy implementation [1] and "permutation" use permutation test with Step-Down method for multiple testing [2]. Tukey HSD method suppose Gaussian model on the aggregated evaluations and permutation test is non-parametric and does not make assumption on the distribution. permutation is the safe choice when the reward is likely to be heavy-tailed or multimodal.
     eval_function: callable or None, default = None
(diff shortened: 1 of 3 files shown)


JulienT01:
tool to download from here : https://github.com/DLR-RM/rl-trained-agents/tree/ca4371d8eef7c2eece81461f3d138d23743b2296

rlberry/manager/comparison.py
```diff
@@ -250,16 +283,49 @@ def compare_agents(
                     ],
                     ignore_index=True,
                 )
-        agent_names = df["agent"].unique()
-        assert len(agent_names) == len(
-            agent_manager_list
-        ), "Each agent must have unique name."
-    else:
-        df = agent_source
```
TimotheeMathieu, line 258: I think this was mistakenly deleted no?

rlberry/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    print("TODO")
+
+
+def import_from_cleanrl():
+    print("TODO")
+
+
+def import_from_hugingface():
+    print("TODO")
```
TimotheeMathieu, line 41: Maybe comment this out for now, so that it does not change the coverage. And anyway if it is not implemented we don't want people to see it in the completion menu of ide.

TimotheeMathieu (commented):
LGTM. Maybe you could add an example in the example folder with download of benchmark + comparison with `compare_agents_data` ? And/or replace the data being compared in the user guide with for the comparison user guide page with data from benchmark?

rlberry/benchmarks/benchmark_utils.py
JulienT01: yes

rlberry/manager/comparison.py
JulienT01: yes bad merge... Now it should be ok !

PR #479, opened 2025-03-21
Replace scikit-fda by handmade Nadaraya-Watson smoothing

TimotheeMathieu:
## Description

In this PR I code by hand the Nadaraya-Watson smoothing. The disadvantage is that I am less sophisticated and I don't unse cross-validation to estimate the bandwidth. The advantage is that we do not depend on scikit-fda anymore.

Still todo:

- [x] add more explanation about bandwidth parameter
- [x] return the smoothed curves as pandas to be able to custom plot them

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -34,8 +34,8 @@ Evaluation and plot
    manager.evaluate_agents
    manager.read_writer_data
    manager.plot_writer_data
-   manager.plot_smoothed_curves
-   manager.plot_synchronized_curves
+   manager.plot_curves_smoothed_NW
+   manager.plot_curves_with_same_x
    manager.compare_agents
 
    manager.tensorboard_to_dataframe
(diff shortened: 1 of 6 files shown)


KohlerHECTOR:
- Make clearer names, e.g. no "ed" for plot_smoothed_curves() maybe call it "smooth_and_plot()".
- Also maybe add the argument to change n_simu column name.
- Refactor code to separate smoothing and plotting ?
- I don't think we need the plot_synchorinized curves.

KohlerHECTOR (approved):
lgtm

KohlerHECTOR (changes requested):
The name of the function are still unclear; what function to use when the sample measurements are asynchronous ?
I tested it and it does not work on my data when there is not the same number of samples measurements across runs.

TimotheeMathieu:
TODO:

- [x] num of simulations may not be a sequence. This implies some bugs. (+ regression.)
- [ ] change name to smooth_NW_and_plot.

